PR #1, opened 2021-10-06
[PR-1134] Learn from errors series - Article on Overlapping instances error

sras:
An article that show some commonly occuring instances of overlapping instances error, and examine their causes and possible fixes.

### DIFF
--- a/learning-from-errors/overlapping-instances/article.md
+++ b/learning-from-errors/overlapping-instances/article.md
@@ -1,10 +1,12 @@
 Not all GHC errors are born equal. Some of them are easy to trace and fix,
-while some of them are not so. And some errors like the 'overlapping
+while some of them are not. And some errors like the 'overlapping
 instances', can have variants that span the entire spectrum. Here we look at
 what this error mean and many of its variants. Along the way, we might learn
 a couple of intersting and advanced things about the behavior of GHC.
 
 ## Simple overlapping instances
+
+Let us look at a basic version of this error that is thrown by the following code.
 
 ```
 
@@ -37,24 +39,23 @@ This gives us the error
 ```
 
 Note that the error happens where there is a call to `printMe` function. If
-there is no call to the function, the error won't be there. ( overlapping
+there is no call to the function, the error won't be there. (overlapping
 instance error is triggered by a function call, and not by an instance
 declaration.  You can put all kinds of overlapping instances in your code, and
 GHC won't bat an eye until you try to call a method in the respective
 typeclass)
 
 Let us see what happens in the call site `printMe (5 ::Int)`.  We have two
-matching instances in scope. The general instance for `Printable a`, and an
+matching instances in scope. The general instance, `Printable a`, and an
 specific instance for `Int`. We call it a 'general instance' because it can
-match any type, while the instance for `Int` can only specifically match the
-`Int` type.
+match any type, while the instance for `Int` can only match the `Int` type.
 
 Here GHC chooses to throw an error, rather than go with the more specific `Int`
 instance. This behavior could help the programmer to not accidentally override a
-general instance. It is easy to spot when both instances are in the same
+general instance by mistake. It is easy to spot when both instances are in the same
 module, but what if the general instance is in another module, or in a
 different package. Silently overriding an existing instance, or not being
-aware of an existing instance could break the program in subtle ways.
+aware of an existing instance while adding a new one could break the program in subtle ways.
 
 ### The Fix
 
@@ -137,7 +138,7 @@ our example, you can use the `OVERLAPS` pragma in either of these instances and
 it will work.
 
 Note that 'OVERLAPPING' pragma just means that it is alright to use that
-instance, if there are other, more general instances, and not an explicit
+instance, even if there are other, more general instances, and not an explicit
 instruction to prefer that instance.
 
 Read more about these pragmas [here](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/pragmas.html#overlapping-overlappable-overlaps-and-incoherent-pragmas)
--- a/learning-from-errors/overlapping-instances/article.md
+++ b/learning-from-errors/overlapping-instances/article.md
@@ -230,6 +230,7 @@ main = fn (5 :: Int)
 fn :: a -> IO ()
 fn x = printMe x
 ```
+</details>
 
 This works, but we see that `fn` is called with an `Int` argument in the `main`
 function. So one can wonder why GHC is not able to figure out that `a` is an
--- a/learning-from-errors/overlapping-instances/article.md
+++ b/learning-from-errors/overlapping-instances/article.md
@@ -233,7 +233,7 @@ fn x = printMe x
 
 This works, but we see that `fn` is called with an `Int` argument in the `main`
 function. So one can wonder why GHC is not able to figure out that `a` is an
-`Int` in this particular call? And they would be right, GHC can, but imagine: if
+`Int` in this particular call? And they would be right, GHC can, but imagine, if
 GHC starts to generate different code for all such polymorphic functions, then
 there will be a lot of copies for a single function if it is called with
 different types.
--- a/learning-from-errors/overlapping-instances/article.md
+++ b/learning-from-errors/overlapping-instances/article.md
@@ -546,7 +546,7 @@ But nevertheless, here is the error:
 ||       In an equation for ‘fn’: fn p a = printMe p a
 ```
 
-Let us look closer at the second instance – just at the instance heads.
+Let us look closer at the second instance.
 
 ```
 instance {-# OVERLAPPING #-} SomeNameClass n (Maybe a) => Printable n (Maybe a) where
--- a/learning-from-errors/overlapping-instances/article.md
+++ b/learning-from-errors/overlapping-instances/article.md
@@ -199,19 +199,11 @@ it cannot pick the most specific instance, causing the error.
 
 ### The Fix
 
-We see that `fn` is called with an `Int` argument in the `main` function. So
-one can wonder why GHC is not able to figure out that `a` is an `Int` in this
-particular call? And they would be right, GHC can, but imagine, if GHC starts
-to generate different code for all such polymorphic functions, then there will
-be a lot of copies for a single function if it is called with different types.
-
-The solution to this problem is nothing other than the plain old typeclass
+The proper solution to this problem is nothing other than the plain old typeclass
 constraints.
 
 So if you add a `Printable a` constraint to `fn`, then the proper instance will
-be passed from the call site, as a hidden argument (a typeclass dictionary),
-and thus the compiler can get away with only generating a single copy of the
-`fn` function.
+be passed from the call site, and the call in `fn` can use this instance.
 
 So that is the proper fix in this situtation. Add a `Printable a` constraint to
 `fn`.
--- a/learning-from-errors/overlapping-instances/article.md
+++ b/learning-from-errors/overlapping-instances/article.md
@@ -231,11 +231,11 @@ fn :: a -> IO ()
 fn x = printMe x
 ```
 
-This works, but we see that `fn` is called with an `Int` argument in `main`
+This works, but we see that `fn` is called with an `Int` argument in the `main`
 function. So one can wonder why GHC is not able to figure out that `a` is an
-`Int` in this particular call? And they would be right, GHC can, but imagine if
+`Int` in this particular call? And they would be right, GHC can, but imagine: if
 GHC starts to generate different code for all such polymorphic functions, then
-the there will be a lot of copies for a single function, if it is called with
+there will be a lot of copies for a single function if it is called with
 different types.
 
 The soloution to this problem is nothing other than the plain old typeclass
@@ -468,7 +468,7 @@ instance Printable a
 instance ... => Printable (f a)
 ```
 
-And it has to pick an instance for the next step, on this information alone.
+And it has to pick an instance for the next step on this information alone.
 In the next step, constraints are matched.
 
 In this example, when the algorithm failed during matching the constraints, if
--- a/learning-from-errors/overlapping-instances/article.md
+++ b/learning-from-errors/overlapping-instances/article.md
@@ -288,7 +288,7 @@ will print "general instance" if you run it.
 
 ## The 'Shortsighted GHC' overlapping instances
 
-Here we look at a variant of this error where one feel that GHC is sometimes
+Here we look at a variant of this error where one feels that GHC is sometimes
 very short sighted.
 
 ```hs
--- a/learning-from-errors/overlapping-instances/article.md
+++ b/learning-from-errors/overlapping-instances/article.md
@@ -415,8 +415,9 @@ if it makese sense.
 
 Don't worry if you have never heard of that name. Because I just made that up.
 
-To demonstrate this we unfortunately need a bit more elaborate setup; So we have
-this code below which throws our beloved error
+To demonstrate this we unfortunately need a bit more elaborate setup, and
+frankly this example is a bit contrived. Anyway, so we have this code below
+which throws our beloved error,
 
 ```
 {-# LANGUAGE DataKinds              #-}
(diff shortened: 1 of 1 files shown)


learning-from-errors/overlapping-instances/article.md
```diff
@@ -0,0 +1,626 @@
+Not all GHC errors are born equal. Some of them are easy to trace and fix,
+while some of them are not. And some errors can have variants that span the entire spectrum.
+
+In this article, we'll look at the `overlapping instances` error. We'll understand the many variants of it and what the error means in each of those cases. Along the way, we might also learn
+a couple of interesting and advanced things about the behavior of GHC.
+
+## Simple overlapping instances
+
+Let us look at a basic version of this error that is thrown by the following code.
```
int-index:
```suggestion
Let us look at a basic version of this error that is triggered by the following code.
```

Throwing an error/exception is a different notion.

learning-from-errors/overlapping-instances/article.md
```diff
@@ -0,0 +1,626 @@
...
+
+```hs
+
+{-# LANGUAGE FlexibleInstances #-}
+
+module Main (main) where
```
int-index: You don’t need the module header, I think it just makes the example longer.

learning-from-errors/overlapping-instances/article.md
```diff
@@ -0,0 +1,626 @@
...
+    instance Printable Int -- Defined at app/Main.hs:11:10
+• In the expression: printMe (5 :: Int)
+```
+
+Note that the error happens where there is a call to `printMe` function. If
+there is no call to the function, the error won't be there.
+
+Overlapping instance error is triggered by a function call, and not by an instance
+declaration.  You can put all kinds of overlapping instances in your code, and
+GHC won't bat an eye until you try to call a method in the respective
```
int-index: This isn’t entirely correct. We don’t need any function calls or methods:

```
{-# LANGUAGE FlexibleInstances #-}

class C a

instance C a
instance C Int

x :: C a => (a, String)
x = (undefined, "Hello")

main = putStrLn (snd x)
```

This code produces the following error:

```
    • Overlapping instances for C a0 arising from a use of ‘x’
      Matching instances:
        instance C a -- Defined at O.hs:5:10
        instance C Int -- Defined at O.hs:6:10
      (The choice depends on the instantiation of ‘a0’
       To pick the first instance above, use IncoherentInstances
       when compiling the other instance declarations)
    • In the first argument of ‘snd’, namely ‘x’
      In the first argument of ‘putStrLn’, namely ‘(snd x)’
      In the expression: putStrLn (snd x)
   |
11 | main = putStrLn (snd x)
   |                      ^
```

But note that `x` is neither a function nor a method of `C`.

So the right way to phrase it would be to talk about _instance declarations_ and _instance search_ (constraint solving). The “overlapping instances” error is not triggered when instances are declared, it is triggered at use sites, when the constraint solver tries to select a particular instance.

The use sites don’t need to be functions or methods.

learning-from-errors/overlapping-instances/article.md
```diff
@@ -0,0 +1,626 @@
...
+  main = printMe (5 :: Int)
+```
+</details>
+
+If you want to mark an instance as being overridable, as well as it being able
+to safely override other instances, you can use the `OVERLAPS` pragme. So in
+our example, you can use the `OVERLAPS` pragma in either of these instances, and
+it will work.
+
+Note that the 'OVERLAPPING' pragma just means that it is alright to use that
```
int-index: Single quotes should be backticks here, I think.

learning-from-errors/overlapping-instances/article.md
```diff
@@ -0,0 +1,626 @@
...
+    (Just x :: Maybe Int) -> putStrLn "general instance for int"
+    Nothing               -> putStrLn "general instance"
+
+main :: IO ()
+main = fn (5 :: Int)
+
+fn :: forall a. Typeable a => a -> IO ()
+fn x = printMe x
+```
+</details>
```
int-index, line 268: I’ve been reading this section in anticipation of the actual fix. I assumed that the `Typeable` story was a tangent, and you were about to propose this change:

```diff
-fn :: a -> IO ()
+fn :: Printable a => a -> IO ()
 fn x = printMe x
```

It doesn’t even require `INCOHERENT`, only `OVERLAPPABLE` or `OVERLAPPING`.

Actually, I would rather see this before you even start discussing `INCOHERENT`, mentioning that GHC’s suggestion is bad.

learning-from-errors/overlapping-instances/article.md
```diff
@@ -0,0 +1,626 @@
...
+And when we re-compile, we get...
+
+```
+• No instance for (Functor MyType) arising from a use of ‘printMe’
+• In the expression: printMe (MyType 'c')
+  In an equation for ‘main’: main = printMe (MyType 'c')
+```
+
+To our great surprise, we find that removing one instance from the two eligible
+instances made GHC reject the remaining instance as well!
```
int-index: This is wrong. The remaining instance isn’t rejected, it is selected quite successfully!

GHC is trying to solve the `Printable (MyType Char)` constraint. It sees that the following instance matches:

```
instance ... => Printable (f a) where
```

That’s it. This instance is selected, with `f=MyType`. Success!

Now as the _next_ stage, GHC starts searching for the instance context, `Functor f`. And the error produced is that the `Functor MyType` instance couldn’t be found, not that the `Printable (f a)` instance was rejected.

int-index: To clarify, I agree that the code is rejected, but I disagree that the instance is rejected.

learning-from-errors/overlapping-instances/article.md
```diff
@@ -0,0 +1,626 @@
...
+come across overlapping instance errors. It is better to carefully examine the
+actual cause of the issue, keeping in mind the algorithm GHC follows to resolve
+instances.
+
+Let us walk through this algorithm and see why the first error happens.
+
+So the very first step is:
+
+```
+Find all instances I that match the target constraint; that is, the target
```
int-index: This is text, it shouldn’t be formatted as code. And the `I` that stands for “instances” should be typeset as math (the syntax is `$ math-goes-here $` IIRC).

learning-from-errors/overlapping-instances/article.md
```diff
@@ -0,0 +1,626 @@
...
+The next step says:
+
+```
+If no candidates remain, the search fails.
+```
+
+Since we have two instances that match, we can continue to the next step, which says,
+
+```
+Eliminate any candidate `IX` for which there is another candidate `IY` such that
```
int-index: Likewise, this is text, not code, so it shouldn’t be monospace. And `IX` and `IY` must be typeset as math.

learning-from-errors/overlapping-instances/article.md
```diff
@@ -0,0 +1,626 @@
...
+without requiring a change to the library.)
+```
+
+We have two candidates, `Printable a` and `Printable (f a)`. Let us process
+`Printable a` first. The rule says to check if there is another candidate `IY`
+such that `IY` is a substitution instance of `IX`. Here `f a` is a
+substituation instance for `a`, because if something can accept `a`, it can accept
+`f a` as well, but not the other way around. So it fits, and the next part of the rule
+says that either `Printable a` is overlappable, or `Printable (f a)` should
+be overlapping. And since this does not match, we cannot eliminate `Printable a`.
```
int-index:
```suggestion
be overlapping. And since this is not the case, we cannot eliminate `Printable a`.
```

learning-from-errors/overlapping-instances/article.md
```diff
@@ -0,0 +1,626 @@
...
+you cannot give `a` to it. Or in other words, `a` is not more specific than `f a`, but
+instead, it is more general.
+
+So after this rule, both instances remain, and the next rule says:
+
+```
+If all the remaining candidates are incoherent, the search succeeds, returning an arbitrary surviving candidate.
+```
+
+None of our instances are marked `Incoherent`, so we proceed to the next rule.
```
int-index:
```suggestion
None of our instances are marked with the `{-# INCOHERENT #-}` pragma, so we proceed to the next rule.
```

int-index:
```suggestion
None of our instances are marked with the `{-# INCOHERENT #-}` pragma, so we proceed to the next rule.
```

learning-from-errors/overlapping-instances/article.md
```diff
@@ -0,0 +1,626 @@
...
+
+We can see that adding `OVERLAPPING` pragma enabled the elimination of the
+instance for `Printable a` at step 3. But the remaining instance `Functor f => Printable (f a)`
+failed to work, but this failure happens at a later phase: when constraints are matched and
+after GHC has picked an instance. This is why we get a `No instance for Functor MyType`
+error instead of an overlapping instance error.
+
+Here we come across a crucial aspect of instance resolution: the algorithm
+never backtracks. When the algorithm failed during matching the constraints, if it could
+backtrack, it could have picked the instance for `Printable a`, which was
```
int-index: There’s another way to phrase this, which you could use even before going through the algorithm.

Instance selection looks at the **instance head** only, and never at the **instance context**. So when instance search takes place, GHC does not see candidate instances like this:

```
instance Printable a
instance Functor f => Printable (f a)
```

It looks at them like this:

```
instance Printable a
instance ... => Printable (f a)
```

And it has to pick an instance based on this information alone.

learning-from-errors/overlapping-instances/article.md
```diff
@@ -0,0 +1,626 @@
...
+```
+Eliminate any candidate `IX` for which there is another candidate `IY` such that
+both of the following hold: `IY` is strictly more specific than `IX`. That is, `IY`
+is a substitution instance of `IX` but not vice versa. Either `IX` is overlappable,
+or `IY` is overlapping. (This “either/or” design, rather than a “both/and”
+design, allow a client to deliberately override an instance from a library
+without requiring a change to the library.)
+```
+
+We have two candidates, `Printable a` and `Printable (f a)`. Let us process
```
int-index, line 373: I never bothered to actually read through that algorithm, thanks for breaking it down step-by-step and applying to a concrete example 👍🏼

learning-from-errors/overlapping-instances/article.md
```diff
@@ -0,0 +1,626 @@
...
+eliminated at a later stage, in favor of the failed instance for `f a`. But instead, the
+algorithm just fails.
+
+### How to fix it?
+
+We can either remove the instance for `f a`, which makes the algorithm pick the
+instance for `Printable a`. Or else we can add a Functor instance for `MyType a`
+if it makes sense.
+
+## Higher-kinded overlapping instances
```
int-index: What about the instances below is “higher-kinded”? Seems like an incorrect use of terminology.

learning-from-errors/overlapping-instances/article.md
```diff
@@ -0,0 +1,626 @@
...
+{-# LANGUAGE PolyKinds              #-}
+{-# LANGUAGE RankNTypes             #-}
+{-# LANGUAGE UndecidableInstances   #-}
+
+module Main (main) where
+
+import Data.Proxy
+import GHC.TypeLits
+
+class Printable n a where
```
int-index, line 482: I think you’re introducing too many things at once. It would be more helpful if you first considered the interaction between overlapping instances and multi-param type classes, and only after that throw `PolyKinds` into the mix.

Also, using a contrived example goes against the spirit of this article, I believe. The first part introduced a simple class `Printable`, the intention behind which was pretty clear, and used this class to explain peculiar GHC behavior. But now you’re just throwing some code together without purpose, so why would I care what GHC has to say about it?

I’m sure it’s possible to find a way to extend `Printable` in a meaningful way. For example, your phantom parameter passed by `Proxy` could be printing configuration

```
class Printable n a where
  printMe :: Proxy n -> a -> IO ()

instance Printable 10 Int where  -- print in base 10
  printMe _ x = print x

instance Printable 16 Int where  -- print in base 16
  printMe _ x = putStrLn (showHex x “”)
```

That’s just an example. But surely there are ways to make the example compelling.

learning-from-errors/overlapping-instances/article.md
NaeosPsy: A quote could perhaps be used, actually.

learning-from-errors/overlapping-instances/article.md
sras: The way I meant this section to work is to first imagine a user observing this error and wondering why GHC is not picking up the general instance to use in `fn`. So I thought the fix should be something that can make it work keeping `fn` free of `Printable a` constraint.

But I think I can include the proper fix at first, and then include the use of `INCOHERENT` after it.

int-index: > why GHC is not picking up the general instance to use in `fn`

Well, it’s not clear that the user would want this behavior. In fact, it would be good to explicitly spell out in the article that there are two reasonable behaviors here:
1.  Just use the default instance (then we need INCOHERENT)
2. Postpone the decision to the use site of `fn` (then we need either `Printable a` or `Typeable a`)

learning-from-errors/overlapping-instances/article.md
sras: I am changing it as follows. I have left the "..reject the remaining instance as well!" to leave it a bit dramatic. Let me know if you think if it's better to remove it.


```
To our great surprise, we find that removing one instance from the two eligible
instances seem to have made GHC reject the remaining instance as well!

But actually, if you actually bothered to look at the error, removing one
instance made GHC proceed with the remaining instance, but only a bit further.
Also the error that is thrown is not an overlapping instance error.

So anyway, we end up in the situation where GHC flip-flops between these two errors.
```

NaeosPsy: I think s/instance/code sounds & delete the new paragraph sounds better if you are not planning to explain the details like Vlad did. It is also the same amount of dramatic.

NaeosPsy: "I just told you a wrong thing" is a very bad pattern in teaching sb.

sras: > I just told you a wrong thing 

Well, it is not a wrong thing. There are two steps in the process by which GHC selects an instance to use at a function call site. AFAIU, both steps should be successful or else the instance cannot be used, and in my description, I consider this to be "rejection" of the instance.

But the "overlapping instance" error that this article discusses, gets thrown only in the first step. So from that point of view, it can be argued that rejection in the second step should not count as "rejection" in this context. This seem to be what Vald is pointing out.

Which is why I think it is not completely wrong, and depends on how you define "rejection" of an instance.

Besides, if we said "removing one instance causes GHC to throw another error", it is actually no big deal (I mean, it happens all the time; you fix one another, GHC moves to another error), and is not at all surprising. It does not convey the impression when someone actually come across this behavior (saying from my own experience). Because both errors are "rejecting" the instance as unsuitable to be used where it is required.

Also FWIW, a bit later, this is included in the same section. So we also explain the details further down..

> But the remaining instance `Functor f => Printable (f a)`
> failed to work, but this failure happens at a later phase: when constraints are matched and
> after GHC has picked an instance. This is why we get a `No instance for Functor MyType`
> error instead of an overlapping instance error.

NaeosPsy: Ok, if you feel like this is better, I trust you. 👌

learning-from-errors/overlapping-instances/article.md
```diff
@@ -0,0 +1,601 @@
...
+
+So this happens because at the call site of `printMe x`, GHC only knows that `x` can
+be of any type, including `Int`. Without knowing if `a` is an `Int` or not,
+it cannot pick the most specific instance, causing the error.
+
+### The Fix
+
+One fix, of course, is to mark the `Int` instance as `INCOHERENT`.  This will
+pick the instance with the information available at the call site, even if a
+different instance is available and could be picked, if more information is available.
```
NaeosPsy:
```suggestion
different instance is available and could be picked if more information were available.
```
Is this perhaps more correct? There is no need for comma before if anyway.

learning-from-errors/overlapping-instances/article.md
```diff
@@ -0,0 +1,601 @@
...
+• In the expression: printMe (5 :: Int)
+```
+
+Note that the error happens where there is a call to `printMe` function. If
+there is no call to the function, the error won't be there.
+
+Overlapping instance error is triggered by an instance search, and not by an
+instance declaration. You can put all kinds of overlapping instances in your
+code, and GHC won't bat an eye until you trigger an instance search, like
+calling a typeclass method.
```
NaeosPsy:
```suggestion
when calling a typeclass method.
```

This seems like it sounds better, and it will be more accurate unless the calling _is_ the instance search.

learning-from-errors/overlapping-instances/article.md
int-index: > if we said "removing one instance causes GHC to throw another error", it is actually no big deal (I mean, it happens all the time; you fix one another, GHC moves to another error)

But that’s exactly what happens in the example :-)

learning-from-errors/overlapping-instances/article.md
sras: I am finding a bit hard to properly address this comment.

>I think you’re introducing too many things at once.

I never meant this article to be exclusively for beginers, and instead I wanted it to be so that people of all levels can find couple of examples that they find interesting.

>using a contrived example goes against the spirit of this article

I agree, but I am unable to find an example that is both sensible and terse enough. Even though we can find a version of `Printable` that is less contrived, shouldn't we also find a version of `SomeNameClass` that make sense?

learning-from-errors/overlapping-instances/article.md
int-index: Also, the article claims that GHC “flip-flops” between the two errors (i.e. switches back and forth), but I see a change only in one direction. We start with an “overlapping instances” error, remove one instance, and get “no Functor instance”. Maybe I’m missing something, but I don’t see that later in the article we’d try another change to the code and it would produce “overlapping instances” again.

NaeosPsy (approved):

learning-from-errors/overlapping-instances/article.md
sras: This "flip-flopping" along with both errors being in the same "class" of error where GHC cannot find an instance to use in the call, is mostly how I see (or used to see) the behavior. 

If it seem that this point of view is not easily relatable, I think we should probably rename/revise the section.

learning-from-errors/overlapping-instances/article.md
sras: Good catch. I think I probably meant "poly-kinded". 

Have made the correction.

learning-from-errors/overlapping-instances/article.md
sras: Have revised the section to..

 
> Overlapping instance error is triggered by an instance search, and not by an
> instance declaration. You can put all kinds of overlapping instances in your
> code, and GHC won't bat an eye until you trigger an instance search, like
> when calling a typeclass method.

learning-from-errors/overlapping-instances/article.md
sras: I have revised this section.

learning-from-errors/overlapping-instances/article.md
sras: Have used a quote and wrapped `I`, `IX` and `IY` in math formatting wrapper.

learning-from-errors/overlapping-instances/article.md
sras: I have renamed and revised this section.

learning-from-errors/overlapping-instances/article.md
sras: Have included this description.

learning-from-errors/overlapping-instances/article.md
```diff
@@ -361,18 +361,9 @@ And when we re-compile, we get...
...
-instance made GHC proceed with the remaining instance, but only a bit further.
-Also the error that results is not an overlapping instance error.
-
-So anyway, we end up in the situation where GHC flip-flops between these two errors.
-
-It shows that it might not be a good idea to blindly remove instances when you
-come across overlapping instance errors. It is better to carefully examine the
-actual cause of the issue, keeping in mind the algorithm GHC follows to resolve
-instances.
+instances seem to have made GHC look closer into the remaining instance, and ended
```
NaeosPsy:
```suggestion
instances seems to have made GHC look closer into the remaining instance, and it ended
```

learning-from-errors/overlapping-instances/article.md
```diff
@@ -361,18 +361,9 @@ And when we re-compile, we get...
...
+up rejecting it after that. It seems that GHC did not look at the instances well
```
NaeosPsy:
```suggestion
up rejecting the instance after that. It seems that GHC did not look at the instances well
```

learning-from-errors/overlapping-instances/article.md
```diff
@@ -361,18 +361,9 @@ And when we re-compile, we get...
...
+enough the first time, before declaring them as redundant.
```
NaeosPsy, line 354:
```suggestion
enough the first time before declaring them as redundant.
```

learning-from-errors/overlapping-instances/article.md
```diff
@@ -286,10 +286,10 @@ ends up being the single 'prime candidate' which gets selected since the
 remaining instance is marked as `INCOHERENT`. This means that the program
 will print "general instance" if you run it.
 
-## Flip-Flopping overlapping instances
+## The 'Shortsighted GHC' overlapping instances
 
-Here we look at a variant of this error that seemingly flip-flops when
-attempts are made to fix it. Let us look at a sample.
+Here we look at a variant of this error where one feel that GHC is sometimes
```
NaeosPsy:
```suggestion
Here we look at a variant of this error where one feels that GHC is sometimes
```

PR #2, opened 2021-10-06
[PR-1091] Article about type inference

Heimdell:


dep-types-inference/dep-types.md
```diff
@@ -0,0 +1,687 @@
...
+  </tr>
+  <tr>
+    <td>
+      <pre>
+        Type
+      </pre>
+    </td>
+    <td>
+      the type of all types (including the `Type` itself)
+    </td>
```
DanielRrr, line 79: Do we have a sort of `TypeInType`? Is it for simplicity? I would emphasise this moment in more detail.

dep-types-inference/dep-types.md
```diff
@@ -0,0 +1,687 @@
...
+
+You also need a mechanism of name refreshing (which will assign some incremented counter to the given name's `uniqueID`) and a scope.
+
+Assuming we have
+
+```haskell
+type Refeshes m :: Constraint
+
+refresh :: Refreshes m => Name -> m Name
+```
```
DanielRrr, line 276: Do you have an implementation of `refresh`? Without that, the section looks a bit incomplete to me.

dep-types-inference/dep-types.md
```diff
@@ -0,0 +1,687 @@
...
+```haskell
+case ... of
+  | pat -> expr  -- those ones
+```
+My solution was to treat them like lambdas - by substituting both pattern and expression with variables captured in the pattern.
+
+Using that, we can return scope-correct programs right off the parser.
+
+_Pro tip learned after a debug session_: Refreshing names of records field declarations in record literal is a bad idea.
+
```
DanielRrr, line 352: To me, it's worth explaining why this is a bad idea in a couple of sentences. The reader might not get your point.

dep-types-inference/dep-types.md
```diff
@@ -0,0 +1,687 @@
...
+## Unification
+
+We can construct the program. Before we're doing inference, we need an auxillary mechanism. It whill compare two types and either find a set of substitutions to convert both to the same type, or bail out with error if that isn't possible. This mechanism is called unification.
+
+Usially, in Hindley-Milner inference, the type of unifier is
+
```
DanielRrr, line 358: It's worth referring to some paper on Hindley-Milner inference to make the post more self-contained.

dep-types-inference/dep-types.md
```diff
@@ -0,0 +1,687 @@
...
+For functions and their types though, it is a bit more complicated process. We need to unify the type of the argument and then replace the argument in the body of function (or dependent function type) with a name from its counterpart and only then we can unify it.
+```haskell
+unify (Pi n t b) (Pi m u c) = do
+  unify t u
+  unify b (subst (Bound m ==> Rigid n) c)
+
+unify (Lam n t b) (Lam m u c) = do
+  unify t u
+  unify b (subst (Bound m ==> Rigid n) c)
+```
```
DanielRrr, line 426: why do you repeat `unify (Lam n t b) (Lam m u c) = do` twice?

dep-types-inference/dep-types.md
```diff
@@ -0,0 +1,687 @@
...
+
+Another intesting thing is record and its type. For record, we unify given signatures of two records, and then their field values. For types of records we only unify types.
+```haskell
+unify (Record ds) (Record gs) = do
+  case zipDecls ds gs of
+    Just quads -> for_ quads \(tn, tm, n, m) -> do
```
DanielRrr, line 432: It's worth saying what `zipDecls` does

dep-types-inference/dep-types.md
```diff
@@ -0,0 +1,687 @@
...
+      unify tn tm  -- unify inferrent types of fields
+      unify  n  m  -- unify the fields themselves
+
+    Nothing -> do
+      die $ Mismatch (Record ds) (Record gs)
+
+unify (Product tds) (Product tgs) = do
+  case zipTDecls tds tgs of
+    Just pairs -> for_ pairs \(n, m) -> do
```
DanielRrr, line 441: The same for `zipTDecls`

dep-types-inference/dep-types.md
```diff
@@ -0,0 +1,687 @@
...
+unify _ LetRec {} = error "unifier: right argument is a let-expression"
+```
+
+And the last case is fired after all other fail, which means that types don't match.
+```haskell
+unify a b = die $ Mismatch a b
+```
+
+## Normalization
+
```
DanielRrr, line 472: You wrote 'Normalization', but you 'normalised' above. Choose your favourite variant for consistency.

dep-types-inference/dep-types.md
```diff
@@ -0,0 +1,687 @@
...
+  <ctx>
+```
+
+The names `Bool`, `True` and `False` will be replaced in `ctx` with (in pseudosyntax)
+
+```
+axiom "True" Bool
+axiom "True" Bool
+axiom "Bool" Type
+```
```
DanielRrr, line 499: Are these something like inference rules for values?

dep-types-inference/dep-types.md
```diff
@@ -0,0 +1,687 @@
...
+```haskell
+infer
+  :: ( Unifies m
+     , HasSubst m
+     , HasContext m
+     , Refreshes m
+     )
+  => AST
+  -> Sem m AST
+```
```
DanielRrr, line 520: What is `Sem`?

dep-types-inference/dep-types.md
```diff
@@ -0,0 +1,687 @@
...
+        tb <- infer b'
+        t' <- unified (eval t) tb
+        return [(n, t')]
+
+      Data n args ctors -> do
+        for_ args \(TDecl _ t) -> do
+          infer t
+
+        let k = telescope args Star
+        check k Star
```
DanielRrr, line 629: Wow, the key place. I would explain what `telescope` is.

DanielRrr: In a couple of words, ofc. There's no need to write a monograph about that.

dep-types-inference/dep-types.md
```diff
@@ -0,0 +1,687 @@
...
+                ret <- getCtorReturnType t
+                ctorCheckReturnType n args ret
+            return [(n', t')]
+        return $ (n, k) : concat nts
+
+    withContext ns do
+      infer b
+```
+
+In case of mutually-recursive definitions, theor bodies (but not the types!) might depend on each other, so we have to bind the constructors to whatever is present in the declarations and do the inference only _after_ we enter the block context.
```
DanielRrr, line 646: their bodies?

dep-types-inference/dep-types.md
```diff
@@ -0,0 +1,687 @@
...
+
+      infer b
+```
+
+There are other cases, but they are either trivial (literals) or the type is contained inside the program directly ("axioms").
+
+## Conclusion
+
+The presence of the typechecker not only allows you to "prevent programs from crashing", but also - if the type system is powerful enough - to partially shift your work onto compiler, by forcing it to implicity calculate some types and, potentially, generate code from them.
+
```
DanielRrr, line 684: I guess the post requires a further reading section.

And mentioning our post written with Vlad :pled:

DanielRrr (approved):

dep-types-inference/dep-types.md
Heimdell: Yes, they are runtime constructor reps, with the type.

dep-types-inference/dep-types.md
Heimdell: An implementation detail - https://hackage.haskell.org/package/polysemy-1.6.0.0/docs/Polysemy.html#t:Sem

dep-types-inference/dep-types.md
Heimdell: Yes, exactly. Otherwise I have to use either additional `Box` or a hierarchy, which requires `Nat`-s and their arithmetics to be somehow present.

dep-types-inference/dep-types.md
Heimdell: I do, will add.

dep-types-inference/dep-types.md
Heimdell:
```haskell
unify (Pi n t b) (Pi m u c) = do
  unify t u
  unify b (subst (Bound m ==> Rigid n) c)
unify (Lam n t b) (Lam m u c) = do
  unify t u
  unify b (subst (Bound m ==> Rigid n) c)
```

Where? The first clause is for `Pi`-types.

Heimdell: They just appear to have the same unification procedure.

PR #3, opened 2021-11-09
[PR-1209] What's That Typeclass: Foldable 

alyoanton9:


alyoanton9:
Not sure about using
````
''
````
instead of Haskell's backticks to use functions as operators
```
``
```
However, I don't know any hacks to use them in "code"

File: whats-that-typeclass-Foldable/article.md [hunk 1]
```diff
@@ -0,0 +1,315 @@
+# What's That Typeclass `Foldable`?
```
heitor-lassarote: Nitpick:
```
# What's That Typeclass: `Foldable`?
```

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,315 @@
...
+
+# What's `Foldable`?
+
+There is a function `reduce` in programming languages, which is appliable to some containers. Its concept is going element by element, applying a given binary function, and keeping the current result as an accumulator, finally returning the value of the accumulator as a summary value of the container. For example, in Python:
+
+```python
+> reduce(lambda x, y: x + y, [1, 2, 3, 4, 5]) # addition of elements in a list
```
heitor-lassarote: I think it would be nice to indicate in a more clear manner here which value is the accumulator (`x`) and which is the element (`y`), perhaps change it to `lambda acc, elem` or similar.

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,315 @@
...
+## `foldr` and `foldl` for a list
+
+To start, let's figure out what are `foldr` and `foldl` for a list. Consider the following type signatures:
+
+```haskell
+foldr :: (a -> b -> b) -> b -> [a] -> b
+foldl :: (b -> a -> b) -> b -> [a] -> b
+```
+
+The `foldr` signature corresponds to the function folding the list from right to left. It takes a function `f` of type `a -> b -> b` as the first argument, the initial (or base) value `z` of type `b` as the second argument and a list `xs = [x_1, x_2, ..., x_n]` of type `[a]` as the third one. Let's introduce the accumulator `acc` which equals to `z` initially. Firstly, `f` is applied to the last element of a list and a base element `z`, the result is being written in an accumulator — `x_n 'f' z`. Then, `f` is applied to the second-to-last element of a list and a current value of an accumulator — `x_(n-1) 'f' (x_n 'f' z)`. After going through all the elements of a  list, `foldr` returns an  accumulator as a summary value, which finally equals to `x_1 'f' (x_2 'f' ... (x_(n-1) 'f' (x_n 'f' z)) ... )`.
```
heitor-lassarote: You can write ``x_n  `f` z`` by using double ticks instead of single ticks, like so: \`\`z_n \`f\` z\`\`.

Also, nitpick: I think it's a bit confusing to use `x_(n-1)` since it's not valid Haskell, perhaps you can use Unicode subscripts: `x₍ₙ₋₁₎` (see [this](https://en.wikipedia.org/wiki/Unicode_subscripts_and_superscripts)), which although still not valid Haskell it's a bit more clear I think? Or use `a`, `b`, ..., `y`, `z` maybe, just so it's a bit easier to read.

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,315 @@
...
+512
+ghci> foldl (^) 2 [2, 3] -- (2 ^ 2) ^ 3
+64
+```
+
+It doesn't matter to the addition if the folding is right-to-left or left-to-right as it's an associative operator. However, when it's folding with exponentiation operator, the order is significant.
+
+## Generalization of `foldr` and `foldl`
+
+Fortunately, it's possible to fold not the lists only! We can implement the instance of `Foldable` whenever a data type has one type argument, in other words, when its [kind](https://wiki.haskell.org/index.php?title=Kind&oldid=62154) is `* -> *`. For instance, a list has one type argument — the type of the list's elements:
```
heitor-lassarote: Not sure if it's better to use `* -> *` or `Type -> Type`. But let's keep it as it is since it's what GHCi shows by default.

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,315 @@
...
+
+`Either a b` data type has two type arguments, corresponding to `Left` and `Right` values, so there is also no reasonable way to fold an `Either a b`:
+
+```haskell
+-- kind of `Either a b`
+Either :: * -> * -> *
+
+-- examples of type `Either a b`
+Left 1 :: Either Int b -- couldn't say what type argument `b` is
+Right "a" :: Either a String -- couldn't say what type argument `a` is
```
heitor-lassarote: Maybe ``can't decide what type `b` is``, or `infer` instead of `decide`, I think it reads a bit more clearly... but see my comment about folding `Either a`.

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,315 @@
...
+
+### Some instances as an example
+
+#### `Maybe a`
+
+Let's take a look at other instances of `Foldable`. One easy-to-understand, but not really useful example is `Maybe a`'s instance. Folding `Nothing` returns a base element, and reducing `Just x` with `foldr f z` (or `foldl f z`) is applying `f` to `x` and `z`:
+
+```haskell
+instance Foldable Maybe where
+    foldr :: (a -> b -> b) -> b -> Maybe a -> b
```
heitor-lassarote: Should we mention `InstanceSigs` here? I don't think it's ideal to show language extensions in a beginner article, but at the same time, we should make sure all presented code compiles. Maybe it would be worth it to create our own `data Option a = None | Some a` and implement `Foldable` for that, or alternatively, comment this part. But this is a nitpick and I don't think it's that important.

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,315 @@
...
+
+```haskell
+foldMap :: (Monoid m, Foldable t) => (a -> m) -> t a -> m
+```
+
+It's important to note that `foldMap` has a [`Monoid`](https://hackage.haskell.org/package/base-4.16.0.0/docs/Prelude.html#t:Monoid) constraint. Its first argument is a function that maps each element of a container into a monoid, the second argument is a container itself. After mapping elements, the results are combined using `(<>)` operator. The order of folding is from right to left, so `foldMap` could be implemented via `foldr`. Let's look how exactly:
+
+```haskell
+foldMap :: (Monoid m, Foldable t) => (a -> m) -> t a -> m
+foldMap f = foldr ((<>) . f) mempty
```
heitor-lassarote: Again, since this is a beginner article, do you think it would perhaps be interesting to not write it in point-free style to keep the code simple?

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,315 @@
...
+```
+
+`foldMap` doesn't have a base element, as only the elements of a container are reduced. However, `foldr` does, so it perfectly makes sense to use an identity of monoid — `mempty`. We can also see  that folding function `f` is composed with `(<>)`, thus, current result is appended to a monoid accumulator on each step.
+
+### Minimal Complete Definition
+
+We have already shown how `foldMap` is implemented using `foldr`. It's not obvious, but `foldr` could also be implemented via `foldMap`! It's off the topic for this article, therefore, please proceed to [the documentation](https://hackage.haskell.org/package/base-4.16.0.0/docs/src/Data.Foldable.html#foldr) for the details. Consequently, to create the `Foldable` instance, you could provide a definition of either `foldr` or `foldMap`, which is exactly the Minimal Complete Definition — `foldMap | foldr`.
```
heitor-lassarote: ~I'm not sure I understand you here, `foldMap | foldr` means you can either implement one or the other. If both were necessary, the minimal complete definition would be `foldMap, foldr`. I think it's important to explain `foldMap`, though, but perhaps reword this part a bit.~ Edit: I'm dumb, I read it again and it's ok.

Also, note that you shouldn't implement `foldMap` in terms of `foldr` and simultaneously `foldr` in terms of `foldMap`, this will just loop forever, so perhaps mention this.

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,315 @@
...
+
+```haskell
+foldr (\s rest -> rest + length s) 0 ["aaa", "bbb", "s"]
+foldr (\rest s -> rest + length s) 0 ["aaa", "bbb", "s"]
+```
+
+<details>
+  <summary>Answer</summary>
+
+The first one
```
heitor-lassarote: Why? Might be interesting to summarize the explanation here (reference the type signature `a -> b -> b`).

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,315 @@
...
+`foldl'` provides forcing to [Weak Head Normal Form](https://wiki.haskell.org/Weak_head_normal_form) in each step, that prevents thunk from being accumulated. So `foldl'` is often a desirable way to reduce a container, especially when it's large, and you want a final strict result.
+
+Here are some "benchmarks", that may vary depending on a computer's characteristics. But they are illustrative enough to grasp the difference between the performance of strict and lazy fold.
+
+```haskell
+-- set GHCI option to show time and memory consuming
+> :set +s
+
+-- lazy `foldl` and `foldr`, length = 10^7
+> foldl (+) 0 [1..10000000]
```
heitor-lassarote: I don't like it that you are talking about BST and then return to lists, maybe move this section about `foldl'` down below the exercises, or mention above that the implementation of `Foldable BinarySearchTree` is an exercise below. At least, as a reader, I was confused wondering what happened to the BST and had to scroll down and up again.

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,315 @@
...
+
+### Implement `foldr` for the `BinarySearchTree a`
+
+<details>
+  <summary>Solution</summary>
+
+```haskell
+foldr :: (a -> b -> b) -> b -> BinarySearchTree a -> b
+foldr _ z Leaf                     = z
+foldr f z (Branch left node right) = foldr f (foldr f (foldr f z right) node) left
```
heitor-lassarote: Perhaps include `instance Foldable BinarySearchTree where` here

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,315 @@
...
+
+It doesn't matter to the addition if the folding is right-to-left or left-to-right as it's an associative operator. However, when it's folding with exponentiation operator, the order is significant.
+
+## Generalization of `foldr` and `foldl`
+
+Fortunately, it's possible to fold not the lists only! We can implement the instance of `Foldable` whenever a data type has one type argument, in other words, when its [kind](https://wiki.haskell.org/index.php?title=Kind&oldid=62154) is `* -> *`. For instance, a list has one type argument — the type of the list's elements:
+
+```haskell
+-- kind of a list
+[] :: * -> *
```
heitor-lassarote, line 110: Not to be confused with the type of a list: `[] :: [a]`. It should be interesting to point that that in GHCi, you can write `:kind []` to see this result, which is different than `:type []`. Since this is the first time a beginner may see this, it's important to clear this up. Punning only makes it harder to explain, but that's how Haskell works. :/

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,315 @@
...
+10000 :: Int
+
+-- kind of `String`
+String :: *
+-- expressions of type `String`
+"" :: String
+"ab" :: String
+```
+
+`Either a b` data type has two type arguments, corresponding to `Left` and `Right` values, so there is also no reasonable way to fold an `Either a b`:
```
heitor-lassarote: You can write `Foldable (Either a)` since `Either a :: * -> *`, so this is not completely true. Instead, I'd say here that the definition for `Foldable (Either a)` (and perhaps `Foldable ((,) a)`) are possible but they are not intuitive since they operate on only one argument.

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,315 @@
...
+-- lazy `foldl` and `foldr`, length = 10^7
+> foldl (+) 0 [1..10000000] 
+50000005000000
+(1.97 secs, 1,612,359,432 bytes)
+
+> foldr (+) 0 [1..10000000] 
+50000005000000
+(1.50 secs, 1,615,360,800 bytes)
+
+-- lazy `foldl` and `foldr`, length = 10^7
```
heitor-lassarote: 10^8

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,315 @@
...
+> foldl (+) 0 [1..100000000]
+*** Exception: stack overflow
+
+> foldr (+) 0 [1..100000000]
+*** Exception: stack overflow
+
+-- strict `foldl'`, length = 10^7..10^9
```
heitor-lassarote: 10^8..10^10

heitor-lassarote (changes requested):

whats-that-typeclass-Foldable/article.md
alyoanton9: ✔️

whats-that-typeclass-Foldable/article.md
alyoanton9: ✔️

whats-that-typeclass-Foldable/article.md
alyoanton9: > You can write x_n `f` z by using double ticks instead of single ticks, like so: ``z_n `f` z``.

Wow! I've replaced 🙂 

I decided to use Unicode subscripts, I believe it looks better and more consistent than `a, b, ..., y, z`.  Furthermore, I hope it will be obvious for a reader, that this is a pseudocode, not Haskell code and so xₙ₋₁ isn't valid.

whats-that-typeclass-Foldable/article.md
alyoanton9: ✔️

whats-that-typeclass-Foldable/article.md
alyoanton9: ✔️

whats-that-typeclass-Foldable/article.md
alyoanton9: I've added some info about `:kind` and `:type` and replaced the comments like `-- (kind|type) of *` with GHCi commands. How does it sound?

whats-that-typeclass-Foldable/article.md
alyoanton9: ✔️

whats-that-typeclass-Foldable/article.md
alyoanton9: ✔️

whats-that-typeclass-Foldable/article.md
alyoanton9: ✔️

whats-that-typeclass-Foldable/article.md
alyoanton9: ✔️

whats-that-typeclass-Foldable/article.md
alyoanton9: ✔️

whats-that-typeclass-Foldable/article.md
alyoanton9: ✔️

whats-that-typeclass-Foldable/article.md
alyoanton9: I mentioned that this is an exercise, and you could implement `instance Foldable BinarySearchTree` yourself

whats-that-typeclass-Foldable/article.md
heitor-lassarote: Thanks, I think it's better.

whats-that-typeclass-Foldable/article.md
heitor-lassarote: I'm still not quite sure about this... What about `\item acc -> f item <> acc`?

whats-that-typeclass-Foldable/article.md
alyoanton9: Oh, I also don't think it's a good idea to introduce language extension. For this particular case, where we just show the definition for library data type, I believe that mentioning type signatures isn't worth. However, when we suggest implementing `instance Foldable BinarySearchTree` as an exercise, it's more likely our reader face with `InstanceSigs`.

So, one option is to write smth like "Bear in mind, that to allow type signature in instance definition, you should use [`InstanceSigs` extension](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/instances.html#extension-InstanceSigs)" in terms of exercise. BTW, I suppose it's not the best link, but others I found were more confusing...

The second option AFAICS is just relying on GHC that will tell a reader that they need to use `InstanceSigs` language extension.

What do you think?

Regarding the custom `Option a` data type, it seems for me that it's redundant to define. The idea here is to show how the `Foldable` instance looks for a certain library data type.

whats-that-typeclass-Foldable/article.md
alyoanton9: Fair, changed ✔️

whats-that-typeclass-Foldable/article.md
heitor-lassarote: There is also a third option: `-- foldr :: (a -> b -> b) -> b -> Maybe a -> b`

Between these options (your suggested ones and commenting), I only have the concern that a beginner may not know how to enable a language extension. So I'm not sure... but I think it's ok to write that warning. This is not something difficult to figure out.

About `Option a`, I agree, let's leave it out.

alyoanton9: > There is also a third option: -- foldr :: (a -> b -> b) -> b -> Maybe a -> b

Yeah, it's possible. But I also agree that it's not difficult indeed to figure out this extension. So, is it okay to leave this comment -- "Bear in mind, that to allow type signature in instance definition, you should use InstanceSigs extension"?

heitor-lassarote: I think that's okay, yeah.

heitor-lassarote:
Thanks for addressing my review. I think it looks quite good!

heitor-lassarote (approved):

alyoanton9:
@heitor-lassarote Waaait, a couple of more things 😄 

I decided to edit exercise part a bit, as it was not pretty to clarify the task in brackets right in ### header. Therefore, I added an expected behaviour for each exercise (except the first one). Could you please take a look at it for last? 🙃

whats-that-typeclass-Foldable/article.md
```diff
@@ -297,14 +297,27 @@ foldl f z (x:xs) =  foldl f (z `f` x) xs
...
+
 <details>
   <summary>Solution</summary>
 
 ```haskell
 instance Foldable BinarySearchTree where
   foldr :: (a -> b -> b) -> b -> BinarySearchTree a -> b
   foldr _ z Leaf                     = z
-  foldr f z (Branch left node right) = foldr f (foldr f (foldr f z right) node) left
+  foldr f z (Branch left node right) = foldr f (foldr f (foldr f z right) [node]) left
```
heitor-lassarote: I think this is more elegant:
```hs
foldr f z (Branch left node right) = foldr f (node `f` foldr f z right) left
```

heitor-lassarote:
Your changes look good to me, just left one comment.

BTW, I think English needs to be reviewed as well, but I don't feel too qualified to comment about it. Maybe @NaeosPsy could take a look at this point?

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,359 @@
...
+
+ghci> :kind String
+String :: *
+ghci> :type ""
+"" :: [Char]
+ghci> :type "ab"
+"ab" :: [Char]
+```
+
+`Either a b` data type has two type arguments corresponding to `Left` and `Right` values, so there is also no reasonable way to fold an `Either a b`. But we could define `instance Foldable (Either a)` since `Either a` has an appropriate `* -> *` kind. The same for 2-tuple data type — `instance Foldable ((,) a)` is possible. However, such instances are not intuitive since they operate on only one type argument.
```
NaeosPsy:
```suggestion
`Either a b` data type has two type arguments corresponding to `Left` and `Right` values, so there is also no reasonable way to fold an `Either a b`. But we could define `instance Foldable (Either a)` since `Either a` has an appropriate `* -> *` kind. The same for 2-tuple data type — `instance Foldable ((,) a)` is possible. However, such instances are not intuitive since they operate on only one of type arguments.
```

NaeosPsy: Might want to call 2-tuple a pair if you want.

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,359 @@
...
+ghci> :type Right 'a'
+Right 'a' :: Either a Char -- can't decide what type argument `a` is
+
+ghci> :kind Either Int
+Either Int :: * -> *
+ghci> :kind (,) Char
+(,) Char :: * -> *
+```
+
+Therefore, for that data types which could have `Foldable` instance, generalized versions of `foldr` and `foldl` have the following type signatures:
```
NaeosPsy: The purpose of Therefore doesn't seem clear, it's just a standalone fact, no?

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,359 @@
...
+1
+```
+
+#### `BinarySearchTree a`
+
+A more interesting and useable `Foldable` instance could be defined for a `BinarySearchTree a` data type. Consider [a Binary Search Tree](https://en.wikipedia.org/w/index.php?title=Binary_search_tree&oldid=1053783914), each node of which either stores a value of type `a` and has left and right subtrees, or is just a leaf without a value. Moreover, for each non-leaf node `n`, all values in its left subtree should be less or equal than the `n`'s value, and all values in its right subtree should be greater than the `n`'s value. 
+
+![Binary Search Tree for the integer numbers](./bst.png)
+
+This conception matches the following Haskell definition:
```
NaeosPsy: What do you mean by "conception" here?

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,359 @@
...
+
+```haskell
+data BinarySearchTree a
+  = Branch (BinarySearchTree a) a (BinarySearchTree a)
+  | Leaf
+```
+
+Imagine that we want to reduce the whole tree to just one value. We could sum values of nodes, multiply them or perform any other binary operation. So that, it's reasonable to define any folding function. We suggest you to implement `Foldable` instance on your own, so it's one of the exercises below. Take into account that to define `foldr`, we need to go through the tree's elements from right to left — from right subtree to left subtree through the non-leaf node connecting them.
```
NaeosPsy: So that, it's reasonable to define any folding function.
 
 This sentence is quite awkward. 

So it's reasonable to define any folding function. 

This is gramatically correct, but still awkward. 

Perhaps this:

It's reasonable to define a folding function that matches your goals.

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,359 @@
...
+
+## Minimal Complete Definition of `Foldable` typeclass
+
+### `foldMap`
+
+The next valuable method of `Foldable` is `foldMap`, having the following declaration:
```
NaeosPsy:
```suggestion
The next important method of `Foldable` is `foldMap`. It has the following declaration:
```

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,359 @@
...
+
+```haskell
+foldMap :: (Monoid m, Foldable t) => (a -> m) -> t a -> m
+```
+
+It's important to note that `foldMap` has a [`Monoid`](https://hackage.haskell.org/package/base-4.16.0.0/docs/Prelude.html#t:Monoid) constraint. Its first argument is a function that maps each element of a container into a monoid, the second argument is a container itself. After mapping elements, the results are combined using `(<>)` operator. The order of folding is from right to left, so `foldMap` could be implemented via `foldr`. Let's look how exactly:
```
NaeosPsy:
```suggestion
It's important to note that `foldMap` has a [`Monoid`](https://hackage.haskell.org/package/base-4.16.0.0/docs/Prelude.html#t:Monoid) constraint. Its first argument is a function that maps each element of a container into a monoid, the second argument is a container itself. After mapping elements, the results are combined using `(<>)` operator. The order of folding is from right to left, so `foldMap` could be implemented via `foldr`. 

Let's look at how exactly that could be done:
```

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,359 @@
...
+
+## Other methods of `Foldable` and things to be careful about when using the type class
+
+`foldl`, `foldr`, and `foldMap` could be considered the core for understanding the nature of `Foldable`. Nonetheless, there are others which are worth mentioning.
+
+### Strict `foldl'`
+
+Haskell uses lazy evaluations by default, that have a positive effect on performance in a lot of cases. Although, the laziness might impact negatively sometimes, and folding is this case exactly. Imagine that you want to sum up the elements in the really huge list. When using `foldl` or `foldr`, a value of an accumulator isn't evaluated on each step, so [thunk](https://wiki.haskell.org/Thunk) is being accumulated. Considering `foldl`, in the first step thunk is just ``z `f` x₁``, in the second — ``(z `f` x₁) `f` x₂``, that's not a big deal. But in the final step, an accumulator stores ``( ... (z `f` x₁) `f` x₂) `f` ...) `f` xₙ₋₁) `f` xₙ``. If the list's length is above ~10^8, thunk becomes too large to store it in a memory, and we get `** Exception: stack overflow`. However, that's not the reason to drop Haskell, as we have a rescue `foldl'`!
```
NaeosPsy: Is each value in the un-evaluated thing a thunk, or is the whole un-evaluated thing a thunk?

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,359 @@
...
+(15.28 secs, 88,000,061,896 bytes)
+
+ghci> foldl' (+) 0 [1..10000000000]
+50000000005000000000
+(263.51 secs, 1,139,609,364,240 bytes)
+```
+
+### Others
+
+There are other methods of `Foldable` type class that could be helpful sometimes. For example, there is a `fold` having `Monoid` constraint, combining elements of a container with `(<>)`. `foldl1` and `foldr1`, that should be used with non-empty containers only. The `length` function, which you might know, is also a member of the `Foldable`, along with `maximum`, `minimum`, and `null`. You might proceed to [the documentation](https://hackage.haskell.org/package/base-4.16.0.0/docs/Data-Foldable.html) to get to know them better.
```
NaeosPsy: "`fold` having `Monoid` constraint" doesn't sound quite right, perhaps "folds a container of Monoids with their <> function" or something?

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,359 @@
...
+
+## Exercises to check yourself
```
NaeosPsy:
```suggestion
## Exercises
```

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,359 @@
...
+
+The theoretical part of our article is over. Hope you know what's that typeclass `Foldable` now! We suggest you to check yourself with our mini exercises.
```
NaeosPsy:
```suggestion
The theoretical part of our article is over. We hope you know what `Foldable` is now! We suggest you to try out your new knowledge with our mini exercises.
```

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,359 @@
...
+
+### Which one definition is correct?
```
NaeosPsy: It seems odd to have task definitions as subheadings, perhaps there is a nicer formatting like lists, or table?

whats-that-typeclass-Foldable/article.md
alyoanton9: I wanted to connect it with the previous explanations of appropriate data types. But if it looks confusing, I'll remove it

whats-that-typeclass-Foldable/article.md
alyoanton9: The conception of Binary Search Tree itself -- how it is structured/defined/represented

whats-that-typeclass-Foldable/article.md
alyoanton9: Thunk is an unevaluated value by definition. Here:

- ``z `f` x₁`` is a thunk. In theory, `z` and `x1` could be unevaluated (thunks) too, but we believe they're evaluated for the sake of simplicity, so it doesn't matter
- ``(z `f` x₁) `f` x₂`` includes 2 thunks -- ``z `f` x₁``  and ``(z `f` x₁) `f` x₂`` itself
- ...

Thus, I guess "the whole un-evaluated thing a thunk" is what you want

whats-that-typeclass-Foldable/article.md
alyoanton9: Does "For example, there is a `fold` that reduces a container of `Monoid`s using `(<>)` function." sound good?

whats-that-typeclass-Foldable/article.md
NaeosPsy: It does look a little bit confusing, yeah.

whats-that-typeclass-Foldable/article.md
NaeosPsy: Yeah then everything is ok

whats-that-typeclass-Foldable/article.md
NaeosPsy: "the <> function", but yeah. 👌

whats-that-typeclass-Foldable/article.md
alyoanton9: [I've replaced](https://github.com/serokell/blog-posts/pull/3/commits/4a1b8c8763718ce93344782165af0272c9d908a1) them with list

whats-that-typeclass-Foldable/article.md
NaeosPsy: I have not heard the word "conception" be used for this purpose. Perhaps "idea" or "structure"?

whats-that-typeclass-Foldable/article.md
alyoanton9: Do you mean that I need "the" article here? Bcs technically `<>` is an operator, but `(<>)` is a function

whats-that-typeclass-Foldable/article.md
NaeosPsy: Seems good!

whats-that-typeclass-Foldable/article.md
NaeosPsy: Yeah, I mean that it needs an article.

whats-that-typeclass-Foldable/article.md
alyoanton9: Yeah, the structure would be better. Sorry, in Russian __conception__ sounds appropriate 😄

alyoanton9:
@NaeosPsy BTW, why "Binary Search Tree", but "weak head normal form"?

NaeosPsy:
> @NaeosPsy BTW, why "Binary Search Tree", but "weak head normal form"?

No reason. They should both be uncapitalized. 🙈

whats-that-typeclass-Foldable/article.md
Martoon-00: Now you say it like if there were no `instance Foldable (Either a)` and `instance Foldable (,) a`, but they in fact exist.

I agree that these instances might be unintuitive, but since they are there, we have to acknowledge them and give some intuition to them. 

(One possible way to explain this: say that elements of the first type argument are ignored, and folding will see only elements of the second type argument, so e.g. `Either l r` will be treated similarly to `Maybe r` by `fold`-like functions)

---

You can also mention, that this can cause bugs when used without caution. E.g. one can write something like `mapM_ verify myTuple` and hope that both elements will be verified (if they are of the same type), but actually only the second one will be checked. 

I once did this mistake, and after that we added a rule to hlint to check that `Foldable` instances for pairs and eithers are never used.

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,364 @@
...
+(15.28 secs, 88,000,061,896 bytes)
+
+ghci> foldl' (+) 0 [1..10000000000]
+50000000005000000000
+(263.51 secs, 1,139,609,364,240 bytes)
+```
+
+### Others
+
+There are other methods of `Foldable` type class that can be helpful sometimes. For example, there is a `fold` that reduces a container of `Monoid`s using the `(<>)` function. `foldl1` and `foldr1`, which should be used with non-empty containers only. The `length` function, which you might know, is also a member of the `Foldable`, along with `maximum`, `minimum`, and `null`. You might proceed to [the documentation](https://hackage.haskell.org/package/base-4.16.0.0/docs/Data-Foldable.html) to get to know them better.
```
Martoon-00: I think, it would be good to say not only that there are `length`, `fold` and other functions, but that usually their implementation is already given based on `foldl` and `foldr` / `foldMap`.

1. This would justify why you considered exactly `foldl` and `foldr` so far;
2. This would also make it clear for the reader, that in real life he needs those two very seldom, more convenient primitives are already provided.
3. This would reveal the full power of `Foldable`, showing that it effectively provides all kinds of basic operations on containers one can remember from other languages (including `for` loops for monadic code, I would mention this here too)

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,364 @@
...
+```haskell
+foldr :: Foldable t => (a -> b -> b) -> b -> t a -> b
+foldl :: Foldable t => (b -> a -> b) -> b -> t a -> b
+```
+
+### Some instances as an example
+
+#### `Maybe a`
+
+Let's take a look at other instances of `Foldable`. One easy-to-understand but not really useful example is `Maybe a`'s instance. Folding `Nothing` returns a base element, and reducing `Just x` with `foldr f z` (or `foldl f z`) is applying `f` to `x` and `z`:
```
Martoon-00: I agree that `foldl` and `foldr` are not that useful for `Maybe`, but saying this for the entire instance is not fully correct because e.g. `for` or `length` over `Maybe` might be pretty helpful occasionally. Perhaps "but not really useful" part could be just stripped.

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,364 @@
...
+3
+ghci> foldr (+) 1 Nothing
+1
+```
+
+#### `BinarySearchTree a`
+
+A more interesting and useable `Foldable` instance could be defined for a `BinarySearchTree a` data type. Consider [a binary search tree](https://en.wikipedia.org/w/index.php?title=Binary_search_tree&oldid=1053783914), each node of which either stores a value of type `a` and has left and right subtrees, or is just a leaf without a value. Moreover, for each non-leaf node `n`, all values in its left subtree should be less or equal than the `n`'s value, and all values in its right subtree should be greater than the `n`'s value. 
+
+![Binary search tree for the integer numbers](./bst.png)
```
Martoon-00: Cool pic :+1:

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,364 @@
...
+
+This structure matches the following Haskell definition:
+
+```haskell
+data BinarySearchTree a
+  = Branch (BinarySearchTree a) a (BinarySearchTree a)
+  | Leaf
+```
+
+Imagine that we want to reduce the whole tree to just one value. We could sum values of nodes, multiply them, or perform any other binary operation. So,it's reasonable to define a folding function that matches your goals. We suggest you try to implement `Foldable` instance on your own; it's one of the exercises below. Take into account that to define `foldr`, we need to go through the tree's elements from right to left — from right subtree to left subtree through the non-leaf node connecting them.
```
Martoon-00:
```suggestion
Imagine that we want to reduce the whole tree to just one value. We could sum values of nodes, multiply them, or perform any other binary operation. So, it's reasonable to define a folding function that matches your goals. We suggest you try to implement `Foldable` instance on your own; it's one of the exercises below. Take into account that to define `foldr`, we need to go through the tree's elements from right to left — from right subtree to left subtree through the non-leaf node connecting them.
```

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,364 @@
...
+foldMap f = foldr (\x acc -> f x <> acc) mempty
+```
+
+`foldMap` doesn't have a base element, as only the elements of a container are reduced. However, `foldr` does, so it perfectly makes sense to use the identity of monoid — `mempty`. We can also see that the folding function `f` is composed with `(<>)`, thus, the current result is appended to a monoid accumulator on each step.
+
+### Minimal Complete Definition
+
+We have already shown how `foldMap` is implemented using `foldr`. It's not obvious, but `foldr` could also be implemented via `foldMap`! It's off-topic for this article, therefore, please proceed to [the documentation](https://hackage.haskell.org/package/base-4.16.0.0/docs/src/Data.Foldable.html#foldr) for the details. Consequently, to create the `Foldable` instance, you cab provide a definition of either `foldr` or `foldMap`, which is exactly the Minimal Complete Definition — `foldMap | foldr`. Note that you shouldn't implement `foldMap` in terms of `foldr` and simultaneously `foldr` in terms of `foldMap` as this will just loop forever. Thus, you implement one of them, and Haskell provides the definitions of all `Foldable`'s methods automatically.
```
Martoon-00:
```suggestion
We have already shown how `foldMap` is implemented using `foldr`. It's not obvious, but `foldr` could also be implemented via `foldMap`! It's off-topic for this article, therefore, please proceed to [the documentation](https://hackage.haskell.org/package/base-4.16.0.0/docs/src/Data.Foldable.html#foldr) for the details. Consequently, to create the `Foldable` instance, you can provide a definition of either `foldr` or `foldMap`, which is exactly the Minimal Complete Definition — `foldMap | foldr`. Note that you shouldn't implement `foldMap` in terms of `foldr` and simultaneously `foldr` in terms of `foldMap` as this will just loop forever. Thus, you implement one of them, and Haskell provides the definitions of all `Foldable`'s methods automatically.
```

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,364 @@
...
+`foldl'` enforces [weak head normal form](https://wiki.haskell.org/Weak_head_normal_form) in each step, thus preventing a thunk from being accumulated. So `foldl'` is often a desirable way to reduce a container, especially when it's large, and you want a final strict result.
+
+Here are some "benchmarks" that may vary depending on a computer's characteristics. But they are illustrative enough to grasp the difference between the performance of strict and lazy fold.
+
+```haskell
+-- set GHCi option to show time and memory consuming
+ghci> :set +s
+
+-- lazy `foldl` and `foldr`, length = 10^7
+ghci> foldl (+) 0 [1..10000000]
```
Martoon-00: Instead of mentioning `length = 10^7` in the comment, why not write `[1..10^7]`?

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,364 @@
...
+
+Let's look at how exactly that could be done:
+
+```haskell
+foldMap :: (Monoid m, Foldable t) => (a -> m) -> t a -> m
+foldMap f = foldr (\x acc -> f x <> acc) mempty
+```
+
+`foldMap` doesn't have a base element, as only the elements of a container are reduced. However, `foldr` does, so it perfectly makes sense to use the identity of monoid — `mempty`. We can also see that the folding function `f` is composed with `(<>)`, thus, the current result is appended to a monoid accumulator on each step.
+
```
Martoon-00, line 242: I would also mention some example of such a `Monoid`, e.g. `Sum`, and briefly illustrate how `foldMap` on it turns out to do exactly the same as the respective `foldr`.

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,364 @@
...
+    ```
+    
+    <details>
+      <summary>Solution</summary>
+    
+    ```haskell
+    instance Foldable BinarySearchTree where
+      foldr :: (a -> b -> b) -> b -> BinarySearchTree a     -> b
+      foldr _ z Leaf                     = z
+      foldr f z (Branch left node right) = foldr f     (node `f` foldr f z right) left
```
Martoon-00: Is this spacing intentional?

Martoon-00 (commented):
Left several comments, but generally - good post :+1:

whats-that-typeclass-Foldable/article.md
alyoanton9: Thanks! Our designer Natalya helped me 😄

whats-that-typeclass-Foldable/article.md
alyoanton9: Would it be appropriate to mention monadic `for` in such a beginner's article? [I've just mentioned](https://github.com/serokell/blog-posts/pull/3/commits/34d47e614d713cb423154df8aefce49507ead369) `maximum`, `minimum`, `length` and `null`, which could be called basic operations on containers, and kind of show the power of `Foldable`.

whats-that-typeclass-Foldable/article.md
alyoanton9: [Done](https://github.com/serokell/blog-posts/pull/3/commits/34d47e614d713cb423154df8aefce49507ead369)

However, I feel like there is a conclusion missing. Is it worth adding smth like "Hope you see now how `foldMap` and `foldr` are connected" or "Hope it's clear now that `foldMap` could be implemented via `foldr`"?

whats-that-typeclass-Foldable/article.md
alyoanton9: No...[Fixed!](https://github.com/serokell/blog-posts/pull/3/commits/34d47e614d713cb423154df8aefce49507ead369)

whats-that-typeclass-Foldable/article.md
Martoon-00: Yeah. I would mention, that `foldMap` and `foldr` are interchangeable, it is just that the former picks combiner and zero point from `Monoid` instance, and the latter accepts them explicitly - or something like that (leaving phrasing of this up to you).

whats-that-typeclass-Foldable/article.md
Martoon-00: Hmm, that's a fair concern. My motivation was that beginners may ask in their heads: "why so complex, where is my old good for-loop?", and we can mention `for` under a sauce of that "Haskell also has `for`, but for more details go check it yourself".

But this is very subjective indeed.

whats-that-typeclass-Foldable/article.md
Martoon-00: And so you can show that `foldMap Sum` is the same as `foldr (+) 0`.

alyoanton9: > And so you can show that foldMap Sum is the same as foldr (+) 0.

Oh, I see that you meant exactly this when wrote:

> e.g. Sum, and briefly illustrate how foldMap on it turns out to do exactly the same as the respective foldr.

Now it seems a bit odd to have these both examples, like one is enough, and `foldMap Sum == foldr (+) 0` looks quite easier than summing squares :/

whats-that-typeclass-Foldable/article.md
alyoanton9: Hmm, okay. I believe, just mentioning it to inform that we also have `for` in our Haskell and provide a link is a good solution.

whats-that-typeclass-Foldable/article.md
alyoanton9: I've come up with idea! We can leave `foldMap Sum == foldr (+) 0`  as an easy example here, and move summing squares to the Exercise part (show implementation using `foldr` and ask to implement via `foldMap` or vice versa).

How does it sound? 🙃

Martoon-00: Yeah, sounds awesome.

alyoanton9: [Pushed](https://github.com/serokell/blog-posts/pull/3/commits/12f3d8dd6f08dc00745bd5f01619e4f99b9210cf). Hope the exercise's solutions are not weird 😅

Martoon-00: Thanks, looks very good to me :+1:

alyoanton9:
@Martoon-00 Seems that all your comments are addressed 🙂 

Could you please approve, if there is nothing more from your side?

Martoon-00 (approved):

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,422 @@
...
+```haskell
+foldr :: Foldable t => (a -> b -> b) -> b -> t a -> b
+foldl :: Foldable t => (b -> a -> b) -> b -> t a -> b
+```
+
+### Some instances as an example
+
+#### `Maybe a`
+
+Let's take a look at other instances of `Foldable`. One easy-to-understand example is `Maybe a`'s instance. Folding `Nothing` returns a base element, and reducing `Just x` with `foldr f z` (or `foldl f z`) is applying `f` to `x` and `z`:
```
NaeosPsy, line 176: Sorry. 😂

Is it "base element" or "accumulator"? Since for every step that is not the first, what you call the base element will be already folded into acc, no?

whats-that-typeclass-Foldable/article.md
```diff
@@ -0,0 +1,422 @@
...
+
+### Minimal complete definition
+
+We have already shown how `foldMap` is implemented using `foldr`. It's not obvious, but `foldr` could also be implemented via `foldMap`! It's off-topic for this article, therefore, please proceed to [the documentation](https://hackage.haskell.org/package/base-4.16.0.0/docs/src/Data.Foldable.html#foldr) for the details. Consequently, to create the `Foldable` instance, you can provide a definition of either `foldr` or `foldMap`, which is exactly the minimal complete definition — `foldMap | foldr`. Note that you shouldn't implement `foldMap` in terms of `foldr` and simultaneously `foldr` in terms of `foldMap` as this will just loop forever. Thus, you implement one of them, and Haskell provides the definitions of all `Foldable`'s methods automatically.
+
+## Other methods of `Foldable` and things to be careful about when using the type class
+
+`foldl`, `foldr`, and `foldMap` could be considered the core for understanding the nature of `Foldable`. Nonetheless, there are others which are worth mentioning.
+
+### Strict `foldl'`
```
NaeosPsy: A bit of a small thing, but foldl' is strict, as I understand, so saying strict foldl` doesn't really make a lot of sense, right?

PR #6, opened 2021-12-06
[PR-1222] Write an article about parser combinators and QQ

heitor-lassarote:
This PR introduces a new blog post about parser combinators. How to make one from scratch, how to use Megaparsec, and as a bonus, how to use quasi quoters to parse in compile time (with anti-quotation support).

### DIFF
--- a/Haskell Parser Combinators/article.md
+++ b/Haskell Parser Combinators/article.md
@@ -23,11 +23,13 @@ This article is composed of three major parts. In the first part, we will implem
 
 In 2001, Daan Leijen and Erik Meijer published a paper titled [Parsec: Direct Style Monadic Parser Combinators For The Real World](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/parsec-paper-letter.pdf), describing the [`parsec`](https://hackage.haskell.org/package/parsec) library, whose design consequently influenced various others, such as [`megaparsec`](https://hackage.haskell.org/package/megaparsec), [`attoparsec`](https://hackage.haskell.org/package/attoparsec), [`trifecta`](https://hackage.haskell.org/package/trifecta), and even libraries outside the Haskell ecosystem, such as [`NimbleParsec`](https://hexdocs.pm/nimble_parsec/NimbleParsec.html) for Elixir, [`parsec`](https://pythonhosted.org/parsec/) for Python, [`FParsec`](https://www.quanttec.com/fparsec/) for F#, among others.
 
-Parser combinators are known to be simple to use without requiring external tools or too many concepts to learn. That is, they are ordinary Haskell functions that can be combined and returned with other parsers. This in turn also makes them idiomatic to use, and being a popular choice among Haskellers, their ecosystem is pretty developed. You should have no trouble finding tutorials and documentation on how to use them.
+Parser combinators are known to be simple to use without requiring external tools or too many concepts to learn. That is, they are ordinary Haskell constructors that can easily be combined and returned with other parsers because of their nature as functions. This in turn also makes them idiomatic to use, and being a popular choice among Haskellers, their ecosystem is pretty developed. You should have no trouble finding tutorials and documentation on how to use them.
 
-It's important to notice that parser combinators also have their flaws. Parser combinators are categorized as LL(∞) parsers, which, elaborating on the points described in Kirill Andreev's [How to Implement an LR(1) Parser](https://serokell.io/blog/how-to-implement-lr1-parser):
+It's important to notice that parser combinators also have their flaws. Parser combinators may be implemented as either LL(1) or LL(∞) parsers, which, elaborating on the points described in Kirill Andreev's [How to Implement an LR(1) Parser](https://serokell.io/blog/how-to-implement-lr1-parser):
 1. Don't support left recursion. That is to say, if there is a rule `foo = foo *> bar`, you might need to refactor your grammar otherwise it will loop forever.
 2. Don't resolve conflicts, which we will see as an example later in this article. This means the ordering of rules can influence the output, and you need to be careful when defining rules that can consume the same input.
+    * In a LL(1) implementation, it will pick the first alternative that maches the input.
+    * In a LL(∞) implementation, it will try the longest possible match, which may be very inneficient.
 3. Many parser combinator libraries will backtrack, which will also be exemplified later, meaning that you might need to be careful to avoid performance penalties.
 4. The backtracking mechanism may, in turn, lead to exponential time complexities.
 
@@ -82,8 +84,6 @@ This is the meaning of each of the type variables:
 * `e`: The type of custom error messages. If we don't have those, we may use `Void` instead.
 * `a`: The result of our parsing function. It represents the structure parsed from our consumed input.
 
-Note that we could implement the parser so it also returns the rest of the input even if there were errors, but to keep the implementation a bit simpler, we'll return it only on success.
-
 Let's begin by creating the most primitive function: `satisfy`. It takes some predicate that operates on the current character, and advances the parser, if possible, returning the consumed character.
 
 We must take care that we are not at the end of the input, however, in which case we should fail.
@@ -129,10 +129,15 @@ instance Functor (Parser i e) where
     case p input of
       Left err -> Left err
       Right (output, rest) -> Right (f output, rest)
-  -- Or:
-  -- fmap f (Parser p) = Parser $ \input -> do
-  --   (output, rest) <- p input
-  --   pure (f output, rest)
+```
+
+Here's a little shorter implementation. Since `Either` is a monad, we can use do-notation.
+
+```hs
+instance Functor (Parser i e) where
+  fmap f (Parser p) = Parser $ \input -> do
+    (output, rest) <- p input
+    pure (f output, rest)
 ```
 
 It's interesting to notice that `Either` is a monad, so we can also write the definition of most things more concisely, as shown in the commented part.
@@ -163,11 +168,18 @@ instance Applicative (Parser i e) where
         case p rest of
           Left err -> Left err
           Right (output, rest') -> Right (f' output, rest')
-  -- Or:
-  -- Parser f <*> Parser p = Parser $ \input -> do
-  --   (f', rest) <- f input
-  --   (output, rest') <- p rest
-  --   pure (f' output, rest')
+```
+
+Or with `Either`'s monad instance:
+
+```hs
+instance Applicative (Parser i e) where
+  pure a = Parser $ \input -> Right (a, input)
+
+  Parser f <*> Parser p = Parser $ \input -> do
+    (f', rest) <- f input
+    (output, rest') <- p rest
+    pure (f' output, rest')
 ```
 
 In our `Monad` instance, we set `return = pure` as usual. Our bind is more interesting: we try to run the given parser with the input, and if it succeeds, we give the produced output (of type `a`) to our continuation `k`, which produces a parser that is executed with the remainder of the output.
@@ -184,10 +196,17 @@ instance Monad (Parser i e) where
           Parser p' = k output
         in
         p' rest
-  -- Or:
-  -- Parser p >>= k = Parser $ \input -> do
-  --   (output, rest) <- p input
-  --   runParser (k output) rest
+```
+
+Again, using the fact that `Either` is a monad, we can simplify this implementation:
+
+```hs
+instance Monad (Parser i e) where
+  return = pure
+
+  Parser p >>= k = Parser $ \input -> do
+    (output, rest) <- p input
+    runParser (k output) rest
 ```
 
 With our `Applicative` and `Monad` instances, we are now able to **parse sequentially**. For example, we defined `char` above to let us parse a single character, but what if we wanted to parse many of them?
@@ -195,16 +214,24 @@ With our `Applicative` and `Monad` instances, we are now able to **parse sequent
 ```hs
 string :: Eq i => [i] -> Parser i e [i]
 string = traverse char
--- Written in a more verbose but perhaps clear way:
--- string :: Eq i => [i] -> Parser i e [i]
--- string [] = pure []
--- string (x : xs) = (:) <$> char x <*> string xs
--- Or even in terms of monad:
--- string [] = return []
--- string (x : xs) = do
---   y <- char x
---   ys <- string xs
---   return (y : ys)
+```
+
+If `traverse` is new to you, it's also equivalent to these definitions:
+
+```hs
+string :: Eq i => [i] -> Parser i e [i]
+string [] = pure []
+string (x : xs) = (:) <$> char x <*> string xs
+```
+
+Or even in terms of `Monad`:
+
+```hs
+string [] = return []
+string (x : xs) = do
+  y <- char x
+  ys <- string xs
+  return (y : ys)
 ```
 
 Let's try it in GHCi:
@@ -472,8 +499,11 @@ We also need an alias for the type which `megaparsec` uses.
 
 ```hs
 type Parser = Parsec
-  Void  -- The type for custom error messages. We have none, so use `Void`.
-  String  -- The input stream type. Let's use `String` for now, but for better performance, you might want to use `Text` or `ByteString`.
+  -- The type for custom error messages. We have none, so use `Void`.
+  Void
+  -- The input stream type. Let's use `String` for now, but for
+  -- better performance, you might want to use `Text` or `ByteString`.
+  String
 ```
 
 ### Booleans
@@ -556,6 +586,13 @@ integer :: Parser Integer
 integer = label "integer" $ read <$> some numberChar
 ```
 
+If you prefer, it's also possible to use `<?>`, which does the same as `label`.
+
+```hs
+integer :: Parser Integer
+integer = read <$> some numberChar <?> "integer"
+```
+
 And now the error message should be slightly improved:
 
 ```hs
@@ -587,7 +624,7 @@ However, later in the tutorial, we will see an even better way that uses Megapar
 
 ### Strings
 
-To parse a string, the idea is to find everything that is between double-quotes. This approach has a weakness, though: we won't be able to parse escaped quotes. Later in this tutorial, we will present a solution to escaped characters.
+To parse a string, the idea is to find everything that is between double-quotes. This approach has a weakness, though: we won't be able to parse escaped quotes. Later in this tutorial, we will present a solution to escaped characters using Megaparsec's lexing mechanisms.
 
 To achieve this, we will use two new functions: `between` and `takeWhileP`. Using `between open close x` is equivalent to `open *> x <* close`, while `takeWhileP` will read as many characters as possible that satisfy some predicate.
 
@@ -657,6 +694,8 @@ sexp = label "S-expression" $ between (char '(') (char ')') ((,) <$> atom <*> ma
 -- Or:
 -- sexp = label "S-expression" $ char '(' *> ((,) <$> atom <*> many atom) <* char ')'
 ```
+
+Note `(,) <$> atom <*> many atom` is pretty similar to `some atom`. [`parser-combinators`](https://hackage.haskell.org/package/parser-combinators-1.3.0/docs/Control-Applicative-Combinators-NonEmpty.html#v:some) exports a non-empty `some` that you can use instead, and have the parser return `Parser (NonEmpty SExp)`. Megaparsec exports the ordinary list version instead, so if you want to use this one, you'll need to import it manually and include a dependency on `parser-combinators`.
 
 After adding it to `atom` with `uncurry SSExp <$> sexp`, we can test it in GHCi:
 
@@ -739,7 +778,7 @@ integer :: Parser Integer
 integer = label "integer" $ lexeme $ read <$> some numberChar
 ```
 
-**Note**: The reader should do the same for `str`, `bool`, and `identifier`.
+**Note**: The reader should do the same for `str`, `bool`, and `identifier`. Alternatively, adding `lexeme` to `atom` will also work.
 
 And now, we can parse S-expressions properly:
 
@@ -884,6 +923,8 @@ And it should fix the problem. However, this is not ideal in terms of performanc
 
 In such a situation this is trivial, but care must be taken in more complex examples, for example, what if we wanted to have `(foo bar)` as well as tuples such as `(foo bar, baz quz)`? The S-expression could be very long and lead to exponential complexity in the worst-case scenario. Recently, a very similar problem [was fixed](https://gitlab.com/morley-framework/morley/-/merge_requests/971) in one of our projects.
 
+As we described in the introduction of this tutorial, it's important to notice Megaparsec normally operates as a LL(1) parser, meaning it tries to look at only one character ahead when executing parsers in an alternative. Using `try` allows us to consume an arbitrary number of characters before backtracking, allowing it to be used as a LL(∞) parser.
+
 In this situation, the best thing to do is to _refactor the grammars to share common parts_. The code above could be written without `try` as such:
 
 ```hs
@@ -948,6 +989,8 @@ numeric = lexeme $ do
         Nothing -> SInteger i
         Just _ -> SDouble $ fromIntegral i
 ```
+
+Note the changes to `str` as well. `manyTill` will run `L.charLiteral` taking characters, including escaped characters until the matching `"` is found.
 
 As a bonus, we got various improvements:
 1. No need to `read` values.
--- a/Haskell Parser Combinators/article.md
+++ b/Haskell Parser Combinators/article.md
@@ -49,7 +49,7 @@ If you instead prefer to jump right into the action, then skip to the [Megaparse
 
 ### Implementation
 
-We will begin by importing the appropriate definitions, on a new file called `Parser.hs`:
+We will begin by importing the appropriate definitions in a new file called `Parser.hs`:
 
 ```hs
 module Parser where
@@ -234,7 +234,7 @@ For `(<|>)`, we want to implement it such that **the first parser to succeed is 
 |Right parser failed|Errors from both parsers|Left parser|
 |Right parser succeeded|Right parser|Left parser|
 
-Here is an implementation of `Alternative`. We constrain `i` and `e` with `Eq` so we can use `nub`, to remove duplicated errors. Note this is not the most efficient or best way to merge errors, but for our purposes, it's good enough.
+Here is an implementation of `Alternative`. We constrain `i` and `e` with `Eq` so we can use `nub` to remove duplicated errors. Note that this is not the best or most efficient way to merge errors, but for our purposes, it's good enough.
 
 ```hs
 instance (Eq i, Eq e) => Alternative (Parser i e) where
@@ -609,7 +609,7 @@ SString "hey"
 
 ### Identifiers
 
-An identifier will be like a variable in a programming language. We can choose the naming convention for such a thing, but here I chose the ones that are frequently used for C-like languages: the first letter must be a letter or an underscore, while the remainder may be a letter, digit, or underscore.
+An identifier will be like a variable in a programming language. We can choose the naming convention for such a thing, but here I chose the one that is frequently used for C-like languages: the first letter must be a letter or an underscore, while the remainder may be a letter, digit, or underscore.
 
 Here we use the `many` function, which is very similar to the `some` function discussed before. The difference is that `any` tries to run the given parser **0 or more times**. In Regex, our `identifier` parser is equivalent to `[a-zA-Z][a-zA-Z0-9]*`.
 
@@ -732,7 +732,7 @@ sexp = label "S-expression" $ lexeme $
   between (lexeme (char '(')) (char ')') ((,) <$> atom <*> many atom)
 ```
 
-You may add `lexeme` in other parsers as well, where it's desirable to skip whitespace after that rule. For instance:
+You may add `lexeme` to other parsers where it's desirable to skip whitespace after the rule. For instance:
 
 ```hs
 integer :: Parser Integer
@@ -769,7 +769,7 @@ expecting S-expression, boolean, identifier, integer, or string
 SInteger 42
 ```
 
-We will create a function that takes our input string, and either returns an error message or the actual result. Such function "bootstraps" our parser with `between skipSpace eof atom`, which will solve the two defects. It will also serve as an auxiliary function that can be called by client code, rather than `parseTest` which is useful for working on GHCi.
+We will create a function that takes our input string, and either returns an error message or the actual result. Such function "bootstraps" our parser with `between skipSpace eof atom`, which will solve the two defects. It will also serve as an auxiliary function that can be called by client code, rather than `parseTest` which is more useful for working in GHCi.
 
 ```hs
 parseSExp :: String -> Either String SExp
--- a/Haskell Parser Combinators/article.md
+++ b/Haskell Parser Combinators/article.md
@@ -84,7 +84,7 @@ This is the meaning of each of the type variables:
 * `e`: The type of custom error messages. If we don't have those, we may use `Void` instead.
 * `a`: The result of our parsing function. It represents the structure parsed from our consumed input.
 
-Let's begin by creating the most primitive function: `satisfy`. It takes some predicate that operates on the current character and advances the parser, if possible, returning the consumed character.
+Let's begin by creating the most primitive function: `satisfy`. It tests the current character with a predicate, and if it succeeds, it will advance the parser and return the consumed character.
 
 We must take care that we are not at the end of the input, however, in which case we should fail.
 
@@ -139,8 +139,6 @@ instance Functor (Parser i e) where
     (output, rest) <- p input
     pure (f output, rest)
 ```
-
-It's interesting to notice that `Either` is a monad, so we can also write the definition of most things more concisely, as shown in the commented part.
 
 Next up is `Applicative`. A good intuition behind `Applicative`s is that `pure` represents an identity element, while `<*>` distributes one argument over the other. Think of this like a multiplication, where `0 * a = a * 0 = 0`, while `1 * a = a * 1 = a`.
 
@@ -304,7 +302,7 @@ Left [Unexpected 'w',Unexpected 'h']
 
 Hopefully, this should give you an intuition of how parser combinators work. Popular parser combinator libraries are more complex than this, as they tend to keep track of extra state for better error messages, optimization, information about line and column, etc., but the overall operations between them are similar.
 
-The complete code for the parser with exercises solutions can be found [here](https://gist.github.com/heitor-lassarote/3e7314956e86b8227f6f6040e69aca9d).
+The complete code for this section with solutions to the exercises below can be found [here](https://gist.github.com/heitor-lassarote/3e7314956e86b8227f6f6040e69aca9d).
 
 ### Exercises:
 1. Change `char` so that it returns errors such as `Left [Expected 'h' 'g']` ("Expected 'h', but got 'g'").
@@ -461,13 +459,13 @@ We will conclude this part here, and if you are interested, proceed to the next 
   * equivalent to JSON: objects, lists, null, numbers, booleans, strings
 -->
 
-We will now take a look at a practical application of parser combinators: parsing S-expressions. Hopefully, this example will give you a solid foundation for creating parsers for your grammars.
+We will now take a look at a practical application of parser combinators: parsing [S-expressions](https://en.wikipedia.org/wiki/S-expressions). Hopefully, this example will give you a solid foundation for creating parsers for your grammars.
 
 Megaparsec is the go-to parser for industrial projects, so we will also use it in our tutorial. If you wish to see how it compares to other parser combinator libraries in Haskell, you can read its [`README.md`](https://github.com/mrkkrp/megaparsec#comparison-with-other-solutions).
 
 Keep in mind you should choose the parser library that better suits your needs. If you plan to parse machine-written data, for example, then Attoparsec may be a better choice. Its interface is similar to Megaparsec's, so the knowledge you will learn here will also be useful.
 
-Without any further ado, we will now present how to use Megaparsec to parse some [`S-expressions`](https://en.wikipedia.org/wiki/S-expressions).
+Without any further ado, we will now present how to use Megaparsec to parse some S-expressions.
 
 ### Representing S-expressions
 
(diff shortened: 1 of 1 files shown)


Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1330 @@
...
+-->
+
+In 2001, Daan Leijen and Erik Meijer published a paper titled [Parsec: Direct Style Monadic Parser Combinators For The Real World](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/parsec-paper-letter.pdf), describing the [`parsec`](https://hackage.haskell.org/package/parsec) library, whose design consequently influenced various others, such as [`megaparsec`](https://hackage.haskell.org/package/megaparsec), [`attoparsec`](https://hackage.haskell.org/package/attoparsec), [`trifecta`](https://hackage.haskell.org/package/trifecta), and even outside the Haskell ecosystem, such as [`NimbleParsec`](https://hexdocs.pm/nimble_parsec/NimbleParsec.html) for Elixir, [`parsec`](https://pythonhosted.org/parsec/) for Python, [`FParsec`](https://www.quanttec.com/fparsec/) for F#, among others.
+
+Parser combinators are known to be simple to use, without requiring external tools or too many concepts to learn. That is, they are ordinary Haskell functions, that can be combined and returned with other parsers. This in turn also makes them idiomatic to use, being a popular choice among Haskellers, having a great ecosystem, and being easy to find examples and documentation.
```
NaeosPsy: The last sentence sounds odd, perhaps you can rephrase it in a simpler way?

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1330 @@
...
+* maybe explain what Alternative is, since beginners may not be familiar with it
+-->
+
+Before we delve into `megaparsec`, it may be interesting to implement our parser combinator from scratch. We will make a simple one that should hopefully help you understand how parser combinators work.
+
+If you instead prefer to jump right into the action, then skip to [Megaparsec tutorial](#megaparsec-tutorial).
+
+### Implementation
+
+We will begin by importing the appropriate definitions, on a new file called `Parser.hs`:
```
NaeosPsy:
```suggestion
We will begin by importing the appropriate definitions in a new file called `Parser.hs`:
```

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1330 @@
...
+I mentioned that `pure` acts as `1`, which is a parser that always succeeds without consuming input. `empty` works like the converse (or `0`), that is a parser that always fails without consuming input.
+
+For `(<|>)`, we want to implement it such that **the first parser to succeed is used**. That is, we first try the left operand, and failing that, we try the second one. The table below shall guide you:
+
+| |Left parser failed|Left parser succeeded|
+|-|-|-|
+|Right parser failed|Errors from both parsers|Left parser|
+|Right parser succeeded|Right parser|Left parser|
+
+Here is an implementation of `Alternative`. We constrain `i` and `e` with `Eq` so we can use `nub`, to remove duplicated errors. Note this is not the most efficient or best way to merge errors, but for our purposes, it's good enough.
```
NaeosPsy:
```suggestion
Here is an implementation of `Alternative`. We constrain `i` and `e` with `Eq` so we can use `nub` to remove duplicated errors. Note that this is not the best or most efficient way to merge errors, but for our purposes, it's good enough.
```

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1330 @@
...
+Finally, in `parse`, use `0` as the initial offset.
+
+Keep in mind it's also possible to operate with line and column when you know you have a function operating in characters.
+
+```hs
+>>> parse (string "hey") "hi"
+Left [Error {erOffset = 1, erError = Expected 'e' 'i'}]
+```
+
+The reader may also choose to instead refactor the parser type to use a custom-defined `State` type that contains the input stream and offset. Many improvements are possible, but we choose to keep simple for didactic purposes.
```
NaeosPsy: I would use the word "educational", "didactic" has some negative connotations.

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1330 @@
...
+  <hr>
+</details>
+
+### Foreword
```
NaeosPsy: Foreword is an introduction, so I would use a different word like "conclusion", "mid-article conclusion", or something that means that without spelling it out. .

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1330 @@
...
+<!--
+* explain why we chose megaparsec
+* parse S-exps
+* make a small format for data parsing, no programming constructs
+  * equivalent to JSON: objects, lists, null, numbers, booleans, strings
+-->
+
+We will now take a look at a practical application with parser combinators: parsing S-expressions. Hopefully, such examples will give you a solid foundation for creating parsers for your grammars.
+
+Megaparsec's [`README.md`](https://github.com/mrkkrp/megaparsec#comparison-with-other-solutions) offers a nice comparison with other popular parser combinator libraries showing its strengths. Being also pretty common in projects, we will also use it for our tutorial.
```
NaeosPsy:
```suggestion
Megaparsec's [`README.md`](https://github.com/mrkkrp/megaparsec#comparison-with-other-solutions) offers a nice comparison with other popular parser combinator libraries showing its strengths.  Since it's pretty common in projects, we will also use it for our tutorial.
```

I would rephrase this and say something like: Megaparsec is the go-to parser for industrial projects, so we will also use it in our tutorial. If you wish to see how it compares to other parser combinator libraries in Haskell, you can read it's README.md 

Otherwise, the start of the paragraph reads a bit confusing.

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1330 @@
...
+```
+
+We will also make a parser to represent our `SExp`-specific variants. As we progress, we will add more parsers to this list.
+
+Here we use `choice`, which is the same as running `<|>` between each element of the list. `choice [a, b, c] = a <|> b <|> c`.
+
+It's important to notice that `choice`, as well as many other useful combinators, are part of the [`parser-combinators`](https://hackage.haskell.org/package/parser-combinators-1.3.0) package, which Megaparsec conveniently re-exports for us. Should you fail to find a specific function in Megaparsec's documentation, chances are that it's defined in `parser-combinators` instead.
+
+```hs
+atom :: Parser SExp
```
NaeosPsy, line 518: I had trouble understanding what this one is at start, perhaps it is possible to spell it out more? 

I think, moving the code underneath "We will also make a parser to represent our `SExp`-specific variants. As we progress, we will add more parsers to this list."

and then writing about choice below code sample would be better.

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1330 @@
...
+In `atom`, include `SString <$> str`.
+
+```hs
+>>> parseTest atom "\"hey\""
+SString "hey"
+```
+
+### Identifiers
+
+An identifier will be like a variable in a programming language. We can choose the naming convention for such a thing, but here I chose the ones that are frequently used for C-like languages: the first letter must be a letter or an underscore, while the remainder may be a letter, digit, or underscore.
```
NaeosPsy:
```suggestion
An identifier will be like a variable in a programming language. We can choose the naming convention for such a thing, but here I chose the one that is frequently used for C-like languages: the first letter must be a letter or an underscore, while the remainder may be a letter, digit, or underscore.
```

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1330 @@
...
+
+Here we use the `many` function, which is very similar to the `some` function discussed before. The difference is that `any` tries to run the given parser **0 or more times**. In Regex, our `identifier` parser is equivalent to `[a-zA-Z][a-zA-Z0-9]*`.
```
NaeosPsy: Here and elsewhere, I don't think there is a need to capitalize regex.

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1330 @@
...
+
+```hs
+>>> parseTest (SBool <$> bool <|> SId <$> identifier) "true"
+SBool True
+
+>>> parseTest (SId <$> identifier <|> SBool <$> bool) "true"
+SId (Identifier {getId = "true"})
+```
+
+In the first case, we parse a boolean before an identifier, which causes `true` to be matched by `bool`. However, in the second example, using `identifier` before `bool` caused `true` to be recognized as an identifier. You may assume that parsers are greedy, and you should be careful while considering such things. Notice that in the second case, the `bool` parser will never be executed.
```
NaeosPsy: Second sentence is a bit awkward 

You may assume that parsers are greedy, and you should be careful while considering such things.

Such things kind of point to the first clause of the sentence (since they are connected), and it raises confusion since I assume you mean such things = order of parsers. 

Perhaps it can go like this:

You should be careful while considering such things. In most cases, you can assume that parsers are greedy. Notice that in the second case..

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1330 @@
...
+```
+
+After adding it to `atom` with `uncurry SSExp <$> sexp`, we can test it in GHCi:
+
+```hs
+>>> parseTest atom "(foo)"
+SSExp (SId (Identifier {getId = "foo"})) []
+```
+
+Ok, looks good so far. What if we have arguments being applied to it?
```
NaeosPsy: Isn't it other way around: what if we apply it to arguments?

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1330 @@
...
+One solution that could come to one's mind is to use the function `sepBy`, which takes a parser and a separator, and use it as ``atom `sepBy` space``. In this case, `space` is a parser defined by Megaparsec that skips zero or more spaces.
+
+```hs
+>>> parseTest (between (char '(') (char ')') $ SSExp <$> atom <*> (space *> atom `sepBy` space)) "(foo bar)"
+SSExp (SId (Identifier {getId = "foo"})) [SId (Identifier {getId = "bar"})]
+```
+
+Of course, this may raise other questions: what if we have a space character before or after the opening and closing parenthesis?
+
+While that works, the [`megaparsec` documentation](https://hackage.haskell.org/package/megaparsec-9.2.0/docs/Text-Megaparsec-Char-Lexer.html) already offers a solution for this kind of problem, namely, a lexer. Indeed, we will see that some of our previous definitions could have been written more elegantly with the help of this module.
```
NaeosPsy: That works = having a space character? Perhaps some rephrasing is necessary.

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1330 @@
...
+
+Now you can use `lexeme` in each of the parsers, and they will automatically skip whitespace after their definitions. Note we also use `lexeme` before our first opening parenthesis, so we can skip space just after the `(`.
+
+```hs
+sexp :: Parser (SExp, [SExp])
+sexp = label "S-expression" $ lexeme $
+  between (lexeme (char '(')) (char ')') ((,) <$> atom <*> many atom)
+```
+
+You may add `lexeme` in other parsers as well, where it's desirable to skip whitespace after that rule. For instance:
```
NaeosPsy:
```suggestion
You may add `lexeme` to other parsers where it's desirable to skip whitespace after the rule. For instance:
```

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1330 @@
...
+1 |   42
+  | ^^^^
+unexpected "  42"
+expecting S-expression, boolean, identifier, integer, or string
+
+>>> parseTest atom "42 foo bar"
+SInteger 42
+```
+
+We will create a function that takes our input string, and either returns an error message or the actual result. Such function "bootstraps" our parser with `between skipSpace eof atom`, which will solve the two defects. It will also serve as an auxiliary function that can be called by client code, rather than `parseTest` which is useful for working on GHCi.
```
NaeosPsy:
```suggestion
We will create a function that takes our input string, and either returns an error message or the actual result. Such function "bootstraps" our parser with `between skipSpace eof atom`, which will solve the two defects. It will also serve as an auxiliary function that can be called by client code, rather than `parseTest` which is more useful for working in GHCi.
```

Or some other changes that show contrast, otherwise just saying that it is useful for ghci sounds a bit odd.

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1330 @@
...
+  pure $ case rightM of
+    Nothing -> SInteger $ read left
+    Just right -> SDouble $ read right
+```
+
+And in `atom`, we simply use `numeric` instead of the two previous parsers.
+
+## Rewriting with Megaparsec's lexer
+
+A lot of the things could be written using the lexer module from Megaparsec. We present the simplified versions below for functions that can be improved.
```
NaeosPsy: I would organize it a little bit, at least the 'improvements" should be below the code. 

Because you said "we present the simplified versions below" and I was reading lines like "where's the code, Lebowski?"

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1330 @@
...
+  = SSExp    SExp [SExp]  -- (foo 0 "hello" bar), (bar (baz 1)), (foo)
+  | SInteger Integer  -- 42
+  | SString  String  -- "hello, world"
+  | SBool    Bool  -- false, true
+  | SId      Identifier  -- foo
+  | SDouble  Double  -- 42f, 3.1415f
+  deriving (Data, Show)
+```
+
+The basic contents for the `TH` module are summarized as follows:
```
NaeosPsy: would link the full file here or somewhere if you have it, since some mistakes with arriving to the final one can appear here, I believe

NaeosPsy (commented):
This must be the most well-written blog post in the history of Serokell. Good job! 👏

Haskell Parser Combinators/article.md
heitor-lassarote: The full file for the TH module is just above this part, it's just a pretty small file, and the complete code for the parser is linked just before the QuasiQuotation part.

Haskell Parser Combinators/article.md
heitor-lassarote: Good point, reorganized and reworded this section a bit.

Haskell Parser Combinators/article.md
heitor-lassarote: I meant explicitly skipping spaces. Reworded it.

Haskell Parser Combinators/article.md
heitor-lassarote: I don't think it is... I meant we have arguments applied to the function. But I reworded it in any case, hopefully, it's a bit more clear now. Or do we apply a function to its arguments? I'm now confused on what is the correct one :confused:

Haskell Parser Combinators/article.md
heitor-lassarote: Reworded it.

Haskell Parser Combinators/article.md
heitor-lassarote: This was a bit surprising to me, but it makes sense, it's just "regular expressions" which shouldn't be capitalized... I've changed it.

Haskell Parser Combinators/article.md
heitor-lassarote: Done! I hope it's better now.

Haskell Parser Combinators/article.md
heitor-lassarote: I've used your suggestion.

Haskell Parser Combinators/article.md
heitor-lassarote: Changed to "conclusion". Seems a bit weird, though...

Haskell Parser Combinators/article.md
heitor-lassarote: I've never heard about that, but I changed it.

Haskell Parser Combinators/article.md
heitor-lassarote: Changed, I hope you like it.

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1332 @@
...
+
+The reader may also choose to instead refactor the parser type to use a custom-defined `State` type that contains the input stream and offset. Many improvements are possible, but we choose to keep it simple for educational purposes.
+  <hr>
+</details>
+
+### Conclusion
+
+In this section, we've learned how to make a simple parsing combinator library. There are still various improvements that can be done. To name a few of them, we can implement more instances, such as `Semigroup` and `Monoid`, `IsString`, `MonadFail`, `MonadPlus`, among others. Error messages can be improved by better keeping track of location (such as line and column), supporting more types such as `Text` and `ByteString`, pretty-printing error messages, etc.
+
+We will conclude this part here, and if you are interested, proceed to the next part, where Megaparsec, an industrial-strength parser combinator library will be introduced with a more practical application rather than our simple examples in this section.
```
NaeosPsy:
```suggestion
We will conclude this part here, and if you are interested, proceed to the next part where Megaparsec, an industrial-strength parser combinator library, will be introduced with a more practical application than our simple examples in this section.
```

Haskell Parser Combinators/article.md
NaeosPsy: Hmm, perhaps something like "What's next?"

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1332 @@
+# Parser Combinators in Haskell
+
+<!--
+* generic introduction
+* say there are four parts for the article
+  1. implement custom parser combinator library
+  2. megaparsec tutorial
+  3. qq
+-->
```
Heimdell, line 7: Thanks for the idea of TODO-comments, I'm stealing it ;)

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1332 @@
...
+
+Parser combinators are known to be simple to use without requiring external tools or too many concepts to learn. That is, they are ordinary Haskell functions that can be combined and returned with other parsers. This in turn also makes them idiomatic to use, and being a popular choice among Haskellers, their ecosystem is pretty developed. You should have no trouble finding tutorials and documentation on how to use them.
+
+It's important to notice that parser combinators also have their flaws. Parser combinators are categorized as LL(∞) parsers, which, elaborating on the points described in Kirill Andreev's [How to Implement an LR(1) Parser](https://serokell.io/blog/how-to-implement-lr1-parser):
```
Heimdell: :blush:

The parsec is _kinda_ LL(1) unless [try](https://hackage.haskell.org/package/parsec-3.1.15.0/docs/Text-Parsec.html#v:try) is used.

Heimdell: And, technically, the combinators themselves are _constructors_ of parsers, the idea is not linked to a particular algorithm.

Here is [Earley parser combinator library](https://hackage.haskell.org/package/Earley), for instance.

They are usually implemented as LL(∞), because it is the easiest algorithm of them all.

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1332 @@
...
+1. Don't support left recursion. That is to say, if there is a rule `foo = foo *> bar`, you might need to refactor your grammar otherwise it will loop forever.
+2. Don't resolve conflicts, which we will see as an example later in this article. This means the ordering of rules can influence the output, and you need to be careful when defining rules that can consume the same input.
```
Heimdell, line 28: Or, rather, autoresolve them to left alternative if LL(1) and longest-match-in-a-most-uneffective-way in LL(∞).

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1332 @@
...
+
+### Implementation
+
+We will begin by importing the appropriate definitions in a new file called `Parser.hs`:
+
+```hs
+module Parser where
+
+import Control.Applicative (Alternative (..))
+import Data.List (nub)
```
Heimdell, line 58: Oh, no.

I understand that this "Set" operation is for sake of simplicity, but eeeeh. I hope it is for `oneOf`, to be used on construction.

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1332 @@
...
+  { runParser :: [i] -> Either [Error i e] (a, [i])
+  }
+```
+
+This is the meaning of each of the type variables:
+* `i`: The input stream. For most cases, we will use `Char`, as `[Char]` is the same as `String`. It's interesting to notice that keeping this as a type variable allows us to parse things that are not strings as well.
+* `e`: The type of custom error messages. If we don't have those, we may use `Void` instead.
+* `a`: The result of our parsing function. It represents the structure parsed from our consumed input.
+
+Note that we could implement the parser so it also returns the rest of the input even if there were errors, but to keep the implementation a bit simpler, we'll return it only on success.
```
Heimdell: That would be interesting. Do you have any links to the strategies to do that, aside from dropping tokens until we find what we was searching for?

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1332 @@
...
+```hs
+data Error i e
+  = EndOfInput  -- Expected more input, but there is nothing
+  | Unexpected i  -- We didn't expect to find this element
+  | CustomError e  -- Extra errors the user may want to create
+  | Empty  -- Used in `Alternative` implementation of `empty`
+  deriving (Eq, Show)
+
+newtype Parser i e a = Parser
+  { runParser :: [i] -> Either [Error i e] (a, [i])
```
Heimdell, line 76: Why not from the transformers? For educational purposes?

Haskell Parser Combinators/article.md
heitor-lassarote: Cool, I like it! I've changed it.

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1332 @@
...
+```hs
+module Parser where
+
+import Control.Applicative (Alternative (..))
+import Data.List (nub)
+```
+
+We will create two data structures that represent our parser.
+
+The first of them represents some error that may happen while parsing. For now, let's have three errors for now: one that is raised when we are expecting input but there is nothing to consume (`EndOfInput`), one when we find an unexpected character (`Unexpected`), and one for any error messages that the user might want to raise (`CustomError`).
```
Martoon-00: Duplicated "for now"

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1332 @@
...
+```hs
+>>> parseTest atom "\"hey\""
+SString "hey"
+```
+
+### Identifiers
+
+An identifier will be like a variable in a programming language. We can choose the naming convention for such a thing, but here I chose the one that is frequently used for C-like languages: the first letter must be a letter or an underscore, while the remainder may be a letter, digit, or underscore.
+
+Here we use the `many` function, which is very similar to the `some` function discussed before. The difference is that `any` tries to run the given parser **0 or more times**. In regex, our `identifier` parser is equivalent to `[a-zA-Z][a-zA-Z0-9]*`.
```
Martoon-00:
```suggestion
Here we use the `many` function, which is very similar to the `some` function discussed before. The difference is that `many` tries to run the given parser **0 or more times**. In regex, our `identifier` parser is equivalent to `[a-zA-Z][a-zA-Z0-9]*`.
```

Martoon-00: Also, underscore seems to be missing in the regex.

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1332 @@
...
+  * Similarly to `quoteExp`, we use `dataToPatQ (const Nothing)` to transform our `SExp` into a `Pat`.
+    * Confusingly, `liftData` is defined as `dataToExpQ (const Nothing)`.
+    * The `const Nothing` can be replaced to handle type-specific cases. We are fine with the default behavior for now but we'll change it later.
+  * Creates a `Pat`, like `[p||]`.
+* `quoteType` creates a Haskell type.
+  * Since we have no `SExp`, `SInteger` etc types matching our constructors, we will fail when someone tries to use our parser as types.
+  * Creates a `Type`, like `[t||]`.
+* `quoteDec` creates a Haskell declaration.
+  * Similarly to `quoteType`, we won't define this one.
+  * Creates a `Pat`, like `[p||]`.
```
Martoon-00: Copy-paste mistake?

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1332 @@
...
+```hs
+>>> let x = SInteger 42 in case [sexp|(foo $x)|] of
+          [sexp|(foo $answer)|] -> print answer
+          _ -> putStrLn "oops"
+SInteger 42
+```
+
+And that's it! You now support quasi-quotations.
+
+The complete code for the parser and quasi-quoter may be found [here](https://gist.github.com/heitor-lassarote/f784c16da80bcfc5271558ede51c70d9).
```
Martoon-00, line 1326: Wow, those two tricks are really cool, I severely lacked them during my development.

Martoon-00 (approved):
Very good post, thanks!

I spotted few minor mistakes, but approving now.

Haskell Parser Combinators/article.md
heitor-lassarote: Removed the second "for now".

Haskell Parser Combinators/article.md
heitor-lassarote: Thanks! I've changed this line.

Haskell Parser Combinators/article.md
heitor-lassarote: Good catch, it was a copy-paste mistake. I fixed it.

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1332 @@
...
+-- string :: Eq i => [i] -> Parser i e [i]
+-- string [] = pure []
+-- string (x : xs) = (:) <$> char x <*> string xs
+-- Or even in terms of monad:
+-- string [] = return []
+-- string (x : xs) = do
+--   y <- char x
+--   ys <- string xs
+--   return (y : ys)
+```
```
Heimdell, line 231: I think, it is worth to write those examples not in comments, but in separate blocks, the article is not in literate haskell format anyway.

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1332 @@
...
+  | SId      Identifier  -- foo
+  deriving (Show)
+```
+
+We also need an alias for the type which `megaparsec` uses.
+
+```hs
+type Parser = Parsec
+  Void  -- The type for custom error messages. We have none, so use `Void`.
+  String  -- The input stream type. Let's use `String` for now, but for better performance, you might want to use `Text` or `ByteString`.
```
Heimdell: I recommend splitting this line on 2 by hand, because otherwise it will be wrapped and wrapper doesn't care about comment indentation.

Heimdell: I does go off-window with horizontal scrolling in Github preview.

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1332 @@
...
+  | ^^^
+unexpected "hey"
+expecting "false", "true", or numeric character
+```
+
+Although the error message is good, wouldn't it better if it said "integer" instead of "numeric character"? For this, we can use **labels**, which allows us to give a name to a parser. To do this, we can simply change our parser like this:
+
+```hs
+integer :: Parser Integer
+integer = label "integer" $ read <$> some numberChar
```
Heimdell, line 583: ... or a `<?>` operator with priority of `0`

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1332 @@
...
+integer = label "integer" $ read <$> (some numberChar <|> ((:) <$> char '-' <*> some numberChar))
+```
+
+However, later in the tutorial, we will see an even better way that uses Megaparsec's own mechanisms.
+  <hr>
+</details>
+
+### Strings
+
+To parse a string, the idea is to find everything that is between double-quotes. This approach has a weakness, though: we won't be able to parse escaped quotes. Later in this tutorial, we will present a solution to escaped characters.
```
Heimdell: There is [charLiteral](https://hackage.haskell.org/package/megaparsec-9.2.0/docs/Text-Megaparsec-Char-Lexer.html#v:charLiteral) that parses both single chars and escape sequences like `\n`.

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1332 @@
...
+
+To achieve this, we will use two new functions: `between` and `takeWhileP`. Using `between open close x` is equivalent to `open *> x <* close`, while `takeWhileP` will read as many characters as possible that satisfy some predicate.
+
+Note that `takeWhileP` also accepts, optionally, a label for each character that it tries to read. For now, we will leave it as `Nothing`.
+
+```hs
+str :: Parser String
+str = label "string" $ between (char '"') (char '"') (takeWhileP Nothing (/= '"'))
```
Heimdell, line 632:
```hs
str = do
  char '"'
  charLiteral `manyTill` char '"'
```

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1332 @@
...
+
+#### S-expressions: The awkward way
+
+The idea here is to parse an atom, followed by 0 or more atoms. Of course, this means that we want to use `many` to parse such atoms. Additionally, an S-expression may be surrounded by parenthesis, so we can use `between` for such a task.
+
+This parser, however, is not as simple as before, but let's first do it the wrong way (which won't work properly), analyze the problem, and come up with a solution to it.
+
+```hs
+sexp :: Parser (SExp, [SExp])
+sexp = label "S-expression" $ between (char '(') (char ')') ((,) <$> atom <*> many atom)
```
Heimdell, line 690: Isn't that `Control.Applicative.some`?

Ah, I get it: nobody made `some` return `Data.List.NotEmpty.NotEmpty` yet.

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1332 @@
...
+```
+
+You may add `lexeme` to other parsers where it's desirable to skip whitespace after the rule. For instance:
+
+```hs
+integer :: Parser Integer
+integer = label "integer" $ lexeme $ read <$> some numberChar
+```
+
+**Note**: The reader should do the same for `str`, `bool`, and `identifier`.
```
Heimdell: ... or on whole atom instead, to not repeat themselves.

Heimdell (approved):
Very good article. Common caveats of LL(1) are discussed, ways out of them are shown.

Haskell Parser Combinators/article.md
heitor-lassarote: For the first point, I added a few more discussions about LL(1) and LL(∞) parsers.

For the second point, I reworded this phrase a bit.

Haskell Parser Combinators/article.md
heitor-lassarote: I've elaborated a bit more on this point. I hope it's better now.

Haskell Parser Combinators/article.md
heitor-lassarote: Later in the tutorial I describe why we use `nub`, it's just for duplicated errors that may arise in some cases.

Haskell Parser Combinators/article.md
heitor-lassarote: I just meant here that our parser could have type `[i] -> (Either [Error i e] a, [i])`, and the changes would just include the "rest" of the output there (or possibly the input again). But I see now it doesn't look like the best idea. I removed this paragraph.

Haskell Parser Combinators/article.md
heitor-lassarote: Do you mean using `ExceptT` or similar here? I'd prefer to use transformers if we had some blog post on the topic first, but yes, let's keep it simple for educational purposes as well.

Haskell Parser Combinators/article.md
heitor-lassarote: I've made separate blocks for them.

Haskell Parser Combinators/article.md
heitor-lassarote: Good point. I've split them so it's shorter now.

Haskell Parser Combinators/article.md
heitor-lassarote: Right, this also exists. I added a point about it.

Haskell Parser Combinators/article.md
heitor-lassarote: I didn't want to present anything from the `Lexer` module just yet since it's discussed later in the tutorial.

Haskell Parser Combinators/article.md
heitor-lassarote: This is how `str` is presented later in the article. But I added another point explaining this part a bit more.

Haskell Parser Combinators/article.md
heitor-lassarote: Actually, there is: https://hackage.haskell.org/package/parser-combinators-1.3.0/docs/Control-Applicative-Combinators-NonEmpty.html#v:some

But Megaparsec doesn't export this function and instead exports the possibly-empty-list one. But I added a small note about it.

Haskell Parser Combinators/article.md
heitor-lassarote: Ah, nice observation! I didn't think of adding it to `atom`. I added a comment about it.

Heimdell (approved):

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1375 @@
...
+  { runParser :: [i] -> Either [Error i e] (a, [i])
+  }
+```
+
+This is the meaning of each of the type variables:
+* `i`: The input stream. For most cases, we will use `Char`, as `[Char]` is the same as `String`. It's interesting to notice that keeping this as a type variable allows us to parse things that are not strings as well.
+* `e`: The type of custom error messages. If we don't have those, we may use `Void` instead.
+* `a`: The result of our parsing function. It represents the structure parsed from our consumed input.
+
+Let's begin by creating the most primitive function: `satisfy`. It takes some predicate that operates on the current character, and advances the parser, if possible, returning the consumed character.
```
NaeosPsy: Is it possible to have a bit simpler structure here?

takes a predicate that operates on the current character, advances the parser (if possible), and returns the consumed character.

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1375 @@
...
+Here's a little shorter implementation. Since `Either` is a monad, we can use do-notation.
+
+```hs
+instance Functor (Parser i e) where
+  fmap f (Parser p) = Parser $ \input -> do
+    (output, rest) <- p input
+    pure (f output, rest)
+```
+
+It's interesting to notice that `Either` is a monad, so we can also write the definition of most things more concisely, as shown in the commented part.
```
NaeosPsy: You mentioned that in the previous line already, sounds a bit awkward now. :D (+ I assume the reference to the "commented" part is now invalid)

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1375 @@
...
+We can see our errors composing with a more complicated expression:
+
+```hs
+>>> runParser ((string "hello" *> string ", globe") <|> string "greetings") "hello, world"
+Left [Unexpected 'w',Unexpected 'h']
+```
+
+Hopefully, this should give you an intuition of how parser combinators work. Popular parser combinator libraries are more complex than this, as they tend to keep track of extra state for better error messages, optimization, information about line and column, etc., but the overall operations between them it similar.
+
+The complete code for the parser with exercises solutions can be found [here](https://gist.github.com/heitor-lassarote/3e7314956e86b8227f6f6040e69aca9d).
```
NaeosPsy: There is something wrong here: 

1. exercises & solutions? 
2. I only see parser code in the gist.

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1375 @@
...
+  * equivalent to JSON: objects, lists, null, numbers, booleans, strings
+-->
+
+We will now take a look at a practical application with parser combinators: parsing S-expressions. Hopefully, such examples will give you a solid foundation for creating parsers for your grammars.
+
+Megaparsec is the go-to parser for industrial projects, so we will also use it in our tutorial. If you wish to see how it compares to other parser combinator libraries in Haskell, you can read its [`README.md`](https://github.com/mrkkrp/megaparsec#comparison-with-other-solutions).
+
+Keep in mind you should choose the parser library that better suits your needs. If you plan to parse machine-written data, for example, then Attoparsec may be a better choice. Its interface is similar to Megaparsec's, so the knowledge you will learn here will also be useful.
+
+Without any further ado, we will now present how to use Megaparsec to parse some [`S-expressions`](https://en.wikipedia.org/wiki/S-expressions).
```
NaeosPsy: I would link in the first paragraph, since people will probably Google when seeing the first occurance. Also, I don't think it needs the code-font.

Haskell Parser Combinators/article.md
```diff
@@ -0,0 +1,1375 @@
...
+  |
+1 | hey
+  | ^^^
+unexpected "hey"
+expecting "false" or "true"
+```
+
+### Integers
+
+To parse integers, one strategy is to parse as many digits (chars whose value is between `'0'` and `'9'`) as possible. This will give us a list of characters, which we can then `read`,
```
NaeosPsy: Something missing here? Or perhaps we just need to change colon to dot. :D

Haskell Parser Combinators/article.md
heitor-lassarote: I reworded a bit, maybe it's easier to understand now.

Haskell Parser Combinators/article.md
heitor-lassarote: Some leftover, I removed this part.

Haskell Parser Combinators/article.md
heitor-lassarote: I reworded a bit, mentioned it contains the solutions to the exercises below.

> I only see parser code in the gist.

I mentioned it's only for this section, which only has the parser code. Or am I missing something?

Haskell Parser Combinators/article.md
heitor-lassarote: I linked in the first paragraph and removed the code-font.

Haskell Parser Combinators/article.md
heitor-lassarote: Eagle eyes, I read this two times and didn't spot it. But now I changed it. Thanks!

NaeosPsy (approved):

Haskell Parser Combinators/article.md
NaeosPsy: I meant that it doesn't have the solutions to the exercises, at least it seems that way to me.

NaeosPsy: My bad, I thought it didn't have the solutions, but it does.

PR #7, opened 2021-12-09
[PR-1268] Follow up article on TTH in GHC 9

heitor-lassarote:
This PR introduces changes to our old Template Haskell articles so they work under GHC 9, as well as a new article describing the changes in the `Code` proposal and how to port existing TTH code from GHC 8 to GHC 9, in both non-compatible and compatible ways.

### DIFF
--- a/Introduction to Template Haskell/article.md
+++ b/Introduction to Template Haskell/article.md
@@ -329,6 +329,7 @@ generateTupleClass size = do
     r = mkName "r"
 
     -- class TupleX t r | t -> r where
+    -- In GHC 9: cDecl = ClassD [] className [PlainTV t (), PlainTV r ()] [FunDep [t] [r]] [mDecl]
     cDecl = ClassD [] className [PlainTV t, PlainTV r] [FunDep [t] [r]] [mDecl]
     --   _X :: t -> r
     mDecl = SigD methodName (AppT (AppT ArrowT (VarT t)) (VarT r))
@@ -340,7 +341,8 @@ For the declaration of the class, we use the `ClassD` constructor with the follo
 
 * `[] :: Ctx` — An empty array of constraints, since we don't impose further restrictions on our types.
 * `className :: Name` — The name of the class being declared.
-* `[PlainTV t, PlainTV r] :: [TyVarBndr ()]` — The bindings of our typeclass, which are `t` and `r`.
+* `[PlainTV t, PlainTV r] :: [TyVarBndr]` — The bindings of our typeclass, which are `t` and `r`.
+  * Note: In GHC 9, use `[PlainTV t (), PlainTV r ()] :: [TyVarBndr ()]` instead.
 * `[FunDep [t] [r]] :: [FunDep]` — A functional dependency of the format `t -> r`.
 * `[mDecl] :: [Dec]` — Our class declarations, containing `_X`, a method of type `t -> r`.
 
--- a/Introduction to Template Haskell/article.md
+++ b/Introduction to Template Haskell/article.md
@@ -134,7 +134,7 @@ What is going on with the `$compose`? This is a _splice_, and it allows us to us
 
 Since Template Haskell is evaluated during compile time, if we replace `x` with a variable that was not bound, such as `z`, we get a `Variable not in scope: z` error instead during the splice generation. Any malformed expressions will be reported like any other regular Haskell expression. In other words, we analyze a Template Haskell declaration as if it was any other ordinary Haskell declaration.
 
-Keep in mind that in GHC 8, the usage of splices may require parentheses or not. For instance, if `compose` came from a qualified import, then you'd need to write `$(Foo.compose)` instead. In GHC 9, the parser is more relaxed and will not require parentheses in such situations.
+Keep in mind that before GHC 9, the usage of splices may require parentheses or not. For instance, if `compose` came from a qualified import, then you'd need to write `$(Foo.compose)` instead. In GHC 9, the parser is more relaxed and will not require parentheses in such situations.
 
 ### Stage restriction
 
--- a/Introduction to Template Haskell/article.md
+++ b/Introduction to Template Haskell/article.md
@@ -134,6 +134,8 @@ What is going on with the `$compose`? This is a _splice_, and it allows us to us
 
 Since Template Haskell is evaluated during compile time, if we replace `x` with a variable that was not bound, such as `z`, we get a `Variable not in scope: z` error instead during the splice generation. Any malformed expressions will be reported like any other regular Haskell expression. In other words, we analyze a Template Haskell declaration as if it was any other ordinary Haskell declaration.
 
+Keep in mind that in GHC 8, the usage of splices may require parentheses or not. For instance, if `compose` came from a qualified import, then you'd need to write `$(Foo.compose)` instead. In GHC 9, the parser is more relaxed and will not require parentheses in such situations.
+
 ### Stage restriction
 
 One caveat of splices exists on where they may be defined and used within some source file. To visualize the problem, we will create a new file called `Main.hs` with the following:
--- a/Introduction to Template Haskell/article.md
+++ b/Introduction to Template Haskell/article.md
@@ -329,8 +329,9 @@ generateTupleClass size = do
     r = mkName "r"
 
     -- class TupleX t r | t -> r where
-    -- In GHC 9: cDecl = ClassD [] className [PlainTV t (), PlainTV r ()] [FunDep [t] [r]] [mDecl]
     cDecl = ClassD [] className [PlainTV t, PlainTV r] [FunDep [t] [r]] [mDecl]
+    -- In GHC 9: cDecl = ClassD [] className [PlainTV t (), PlainTV r ()] [FunDep [t] [r]] [mDecl]
+
     --   _X :: t -> r
     mDecl = SigD methodName (AppT (AppT ArrowT (VarT t)) (VarT r))
 ```
--- a/Typed Template Haskell/article.md
+++ b/Typed Template Haskell/article.md
@@ -1,4 +1,4 @@
-Welcome to our second post on Template Haskell! 
+Welcome to our second post on Template Haskell!
 
 Today we will take a quick look at typed Template Haskell. This article assumes some familiarity with Template Haskell (TH) already. If this is your first journey with TH, then check out our [introduction to Template Haskell](https://serokell.io/blog/introduction-to-template-haskell) first.
 
@@ -68,7 +68,9 @@ SigE (LitE (IntegerL 42)) (ConT GHC.Base.String)
 
 ## Typed splices
 
-Just like we had untyped splices such as `$foo`, now we also have typed splices, written as `$$foo`. Note, however, that if your GHC version is below 9.0, you may need to write `$$(foo)` instead.
+Just like we had untyped splices such as `$foo`, now we also have typed splices, written as `$$foo`. In this section, we will see how to define and splice typed Template Haskell code.
+
+We reiterate one point mentioned in our previous article: in GHC 8, the usage of splices may require parentheses or not. For instance, if `foo` came from a qualified import, then you'd need to write `$$(Foo.foo)` instead. In GHC 9, the parser is more relaxed and will not require parentheses in such situations.
 
 ## Example: calculating prime numbers
 
@@ -267,7 +269,7 @@ Typed Template Haskell may have some difficulties resolving overloads. Surprisin
 ... mempty' = [|| mempty ||]
 
 >>> x :: String
-... x = id $$(mempty')
+... x = id $$mempty'
 <interactive>:549:11: error:
     • Ambiguous type variable ‘a0’ arising from a use of ‘mempty'’
       prevents the constraint ‘(Monoid a0)’ from being solved.
@@ -279,8 +281,8 @@ Typed Template Haskell may have some difficulties resolving overloads. Surprisin
         ...plus 7 others
         (use -fprint-potential-instances to see them all)
     • In the expression: mempty'
-      In the Template Haskell splice $$(mempty')
-      In the first argument of ‘id’, namely ‘$$(mempty')’
+      In the Template Haskell splice $$mempty'
+      In the first argument of ‘id’, namely ‘$$mempty'’
 ```
 
 Annotating `mempty'` may resolve it in this case:
--- a/Typed Template Haskell/article.md
+++ b/Typed Template Haskell/article.md
@@ -70,7 +70,7 @@ SigE (LitE (IntegerL 42)) (ConT GHC.Base.String)
 
 Just like we had untyped splices such as `$foo`, now we also have typed splices, written as `$$foo`. In this section, we will see how to define and splice typed Template Haskell code.
 
-We reiterate one point mentioned in our previous article: in GHC 8, the usage of splices may require parentheses or not. For instance, if `foo` came from a qualified import, then you'd need to write `$$(Foo.foo)` instead. In GHC 9, the parser is more relaxed and will not require parentheses in such situations.
+We reiterate one point mentioned in our previous article: before GHC 9, the usage of splices may require parentheses or not. For instance, if `foo` came from a qualified import, then you'd need to write `$$(Foo.foo)` instead. In GHC 9, the parser is more relaxed and will not require parentheses in such situations.
 
 ## Example: calculating prime numbers
 
@@ -330,7 +330,7 @@ Furthermore, as discussed, the [caveat](#caveat) in the previous section doesn't
 ""
 ```
 
-Should you need to explicitly convert between the old and new types, you may pattern match directly on `Code` or use the following functions:
+Should you need to explicitly convert between the old and new types, you may pattern match directly on `Code` or use the following functions from [Language.Haskell.TH](https://hackage.haskell.org/package/template-haskell-2.18.0.0/docs/Language-Haskell-TH.html#t:Code):
 
 ```hs
 liftCode :: m (TExp a) -> Code m a
--- a/Typed Template Haskell/article.md
+++ b/Typed Template Haskell/article.md
@@ -2,7 +2,7 @@ Welcome to our second post on Template Haskell!
 
 Today we will take a quick look at typed Template Haskell. This article assumes some familiarity with Template Haskell (TH) already. If this is your first journey with TH, then check out our [introduction to Template Haskell](https://serokell.io/blog/introduction-to-template-haskell) first.
 
-For this article, we will be using GHC 8.10.4.
+For this article, we will be using GHC 8.10.4. However, we will also show the changes that are necessary so the code works in GHC 9.
 
 ## Why typed TH?
 
@@ -110,6 +110,8 @@ Keep in mind that these functions are _very_ inefficient, so make sure to use a 
 
 Now for our Template Haskell version. As usual, let's create two files, `TH.hs` and `Main.hs`, to work with through this example.
 
+Please note that for GHC 9, you will need to write `Code Q a` in places where we write `Q (TExp a)`. For more details, see the section about [Compatibility with GHC 9](#compatibility-with-ghc-9).
+
 This is what should be in `TH.hs`:
 ```hs
 {-# LANGUAGE TemplateHaskell #-}
@@ -121,6 +123,7 @@ import Language.Haskell.TH
 import Primes (isPrime)
 
 primesUpTo' :: Integer -> Q (TExp [Integer])
+--primesUpTo' :: Integer -> Code Q [Integer]
 primesUpTo' n = go 2
   where
     go i
@@ -179,6 +182,7 @@ Had we made any mistakes in the definition, for example, by using the following 
 
 ```hs
 primesUpTo' :: Integer -> Q (TExp [Integer])
+--primesUpTo' :: Integer -> Code Q [Integer]
 primesUpTo' n = go 2
   where
     go i
@@ -255,6 +259,7 @@ And the corresponding TH function as:
 
 ```hs
 primesUpTo' :: Integer -> Q (TExp [Integer])
+--primesUpTo' :: Integer -> Code Q [Integer]
 primesUpTo' n = [|| primesUpTo n ||]
 ```
 
@@ -262,7 +267,7 @@ And with this, you should be ready to use typed Template Haskell in the wild.
 
 ## Caveat
 
-Typed Template Haskell may have some difficulties resolving overloads. Surprisingly, the following does not type-check:
+Typed Template Haskell may have some difficulties resolving overloads in versions older than GHC 9. Surprisingly, the following does not type-check:
 
 ```hs
 >>> mempty' :: Monoid a => Q (TExp a)
@@ -296,6 +301,43 @@ Annotating `mempty'` may resolve it in this case:
 ```
 
 An [open ticket](https://gitlab.haskell.org/ghc/ghc/-/issues/10271) exists describing the issue, but if you run into some strange errors, it's a good idea to keep it in mind.
+
+## Compatibility with GHC 9
+
+In GHC 9, the types used by typed Template Haskell were changed according to the [Make Q (TExp a) into a newtype](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0195-code-texp.rst) proposal.
+
+This proposal changed the typed expression quasi-quoter so it now returns `Quote m => Code m a` instead of the old `Quote m => m (TExp a)` from [`template-haskell-2.17.0.0`](https://hackage.haskell.org/package/template-haskell-2.17.0.0/docs/Language-Haskell-TH.html#t:Code).
+
+<!-- TODO: Write new article, add link to it, uncomment this part.
+If the reader is curious, please take a moment to read our []() article detailing a migration guide of typed TH code to GHC 9, where we discuss a backward-compatible way to integrate typed Template Haskell code in both GHC 8 and GHC 9.
+-->
+
+Thankfully, changing the examples to work with GHC 9 is not difficult. In short, you may replace the parts like `Q (TExp a)` with `Quote m => Code m a` (or `Code Q a`) and the code should compile.
+
+```hs
+primesUpTo' :: Quote m => Integer -> Code m [Integer]
+primesUpTo' n = [|| primesUpTo n ||]
+```
+
+Furthermore, as discussed, the [caveat](#caveat) in the previous section doesn't apply anymore to GHC 9, and you will be able to write the example above.
+
+```hs
+>>> mempty' :: Monoid a => Code Q a
+... mempty' = [|| mempty ||]
+
+>>> x :: String
+... x = id $$mempty'
+
+>>> x
+""
+```
+
+Should you need to explicitly convert between the old and new types, you may pattern match directly on `Code` or use the following functions:
+
+```hs
+liftCode :: m (TExp a) -> Code m a
+examineCode :: Code m a -> m (TExp a)
+```
 
 ## Further reading
 
--- a/Typed Template Haskell/article.md
+++ b/Typed Template Haskell/article.md
@@ -306,9 +306,9 @@ An [open ticket](https://gitlab.haskell.org/ghc/ghc/-/issues/10271) exists descr
 
 In GHC 9, the types used by typed Template Haskell were changed according to the [Make Q (TExp a) into a newtype](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0195-code-texp.rst) proposal.
 
-This proposal changed the typed expression quasi-quoter so it now returns `Quote m => Code m a` instead of the old `Quote m => m (TExp a)` from [`template-haskell-2.17.0.0`](https://hackage.haskell.org/package/template-haskell-2.17.0.0/docs/Language-Haskell-TH.html#t:Code).
+This proposal changed the typed expression quasi-quoter so it now returns `Quote m => Code m a` instead of the old `Q (TExp a)` which was used prior to GHC 9.
 
-If the reader is curious, please take a moment to read our [Typed Template Haskell in GHC 9](https://serokell.io/blog/typed-template-haskell-in-ghc-9) article detailing a migration guide of typed TH code to GHC 9, where we discuss a backward-compatible way to integrate typed Template Haskell code in both GHC 8 and GHC 9.
+If you're curious, please take a moment to read our [Typed Template Haskell in GHC 9](https://serokell.io/blog/typed-template-haskell-in-ghc-9) article detailing a migration guide of typed TH code to GHC 9, where we discuss a backward-compatible way of integrating typed Template Haskell code in both GHC 8 and GHC 9.
 
 Thankfully, changing the examples to work with GHC 9 is not difficult. In short, you may replace the parts like `Q (TExp a)` with `Quote m => Code m a` (or `Code Q a`) and the code should compile.
 
--- a/Typed Template Haskell/article.md
+++ b/Typed Template Haskell/article.md
@@ -306,7 +306,7 @@ An [open ticket](https://gitlab.haskell.org/ghc/ghc/-/issues/10271) exists descr
 
 In GHC 9, the types used by typed Template Haskell were changed according to the [Make Q (TExp a) into a newtype](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0195-code-texp.rst) proposal.
 
-This proposal changed the typed expression quasi-quoter so it now returns `Quote m => Code m a` instead of the old `Q (TExp a)` which was used prior to GHC 9.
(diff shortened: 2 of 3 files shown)


heitor-lassarote:
@NaeosPsy, there is a [TODO](https://github.com/serokell/blog-posts/pull/7/files#diff-5785bb9f39abf508ac66467582b49784454ee716fd95cacc6785dd2cbe41b714R311-R313) that should be changed before the new article is posted. If you know what the link is, can you either change it or tell me what it is so I can uncomment this part and add the link?

NaeosPsy:
> @NaeosPsy, there is a [TODO](https://github.com/serokell/blog-posts/pull/7/files#diff-5785bb9f39abf508ac66467582b49784454ee716fd95cacc6785dd2cbe41b714R311-R313) that should be changed before the new article is posted. If you know what the link is, can you either change it or tell me what it is so I can uncomment this part and add the link?

I think  https://serokell.io/blog/typed-template-haskell-in-ghc-9 is ok, I've created an unpublished post with that slug on website.

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,265 @@
+# Typed Template Haskell in GHC 9
+
+Welcome to our third post about Template Haskell! Today we will take a look at the changes that were made in GHC 9 regarding Typed Template Haskell (TTH) and how to use the [`th-compat`](https://hackage.haskell.org/package/th-compat) library to make TTH code that will work with both GHC 8 and GHC 9.
+
+<!-- TODO: Actually amend the article! -->
```
int-index: The TODO is completed by this patch, right?

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,265 @@
...
+In our [previous blog post](https://serokell.io/blog/typed-template-haskell-overview), we made an overview of Template Haskell in GHC 8. The article was later amended with the changes required so that the examples compile in GHC 9 in a non-backward compatible way. Make sure to read that post before you continue!
+
+## Changed Typed Template Haskell specification
+
+The `template-haskell` package was changed in version 2.17.0.0 and GHC version 9.0 according to the [Make Q (TExp a) into a newtype](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0195-code-texp.rst) proposal, in which the typed expression quasi-quoter (`[|| ... ||]`) now returns a different datatype.
+
+Under the new specification, typed expressions now return `Quote m => Code m a` instead of `Q (TExp a)`. This is a breaking change for existing code, and codebases targetting GHC 9.0 either need to adapt their TTH code or use an alternative approach for compatibility.
```
int-index:
```suggestion
Under the new specification, typed quotations now return `Quote m => Code m a` instead of `Q (TExp a)`. This is a breaking change for existing code, and codebases targeting GHC 9.0 either need to adapt their TTH code or use an alternative approach for compatibility.
```

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,265 @@
...
+
+The `Code` newtype is simply a wrapper around an `m (TExp a)` (plus some levity-polymorphism mechanisms), defined as such:
+
+```hs
```
int-index, line 23: You could start with a simpler definition

```
newtype Code m a = Code (m (TExp a))
```

This will give the reader a good enough understanding. And after that give the full definition for completeness.

int-index (commented):
The content of the article is fine, but the style does not match the context.

Imagine a person facing TH type errors when migrating to GHC 9.0

Would they want to read an entire article like this, with all the prose and code examples? No.

They’d want short facts and steps. Preferably the entire migration should fit in a paragraph or two. Otherwise I foresee annoyed readers trying to dig through all text, scrolling chaotically and thinking “just tell me what to do!”

Also, you have a 10-line long function `logLevelFromFlag` that you use as your example. Then you make a small adjustment to it and copy the entire definition – is the reader expected to diff them in their head by reading carefully?

Could we show the diff instead of the entire snippet, like this?

```diff
- logLevelFromFlag :: Q (TExp LogLevel)
+ logLevelFromFlag :: Code Q LogLevel
```

typed-th-in-ghc-9/article.md
heitor-lassarote: Yes, that's right. I've removed this comment.

typed-th-in-ghc-9/article.md
heitor-lassarote: Oops, thanks!

typed-th-in-ghc-9/article.md
heitor-lassarote: Good idea, I hope it's a bit more clear now.

heitor-lassarote:
@int-index, I've changed it to diffs, I hope it's better now. However, I chose to keep the entire snippet of the function, as overall there are only a few unchanged lines and it looks clearer like this to me.

I've also deleted some information that is more like an extra to the guide and joined some relevant paragraphs when possible. I can also go further and only show how to use `th-compat` (i.e. skip the `Code` part), or delete even more parts if you believe it's better. My feeling is that the post is not too long so there shouldn't be so much confusion. Pray tell me if you think it can be further trimmed.

int-index (approved):
Much clearer with diffs, thanks! I hope our blog engine can highlight them appropriately

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,205 @@
...
+
+For example, if we had imported `logLevelFromFlag` qualified, then you'd write `$$(TH.logLevelFromFlag)` in GHC 8, otherwise you'd get a parser error:
+
+```hs
+>>> $$TH.logLevelFromFlag
+
+<interactive>:200:1: error: parse error on input ‘$$’
+```
+
+But it works otherwise in GHC 9:
```
NaeosPsy:
```suggestion
But it works in GHC 9:
```

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,205 @@
...
+
+```hs
+>>> $$TH.logLevelFromFlag
+Production
+```
+
+## Conclusion
+
+In this post, we've seen the changes introduced in GHC 9 regarding Typed Template Haskell. We learned about two different ways to migrate typed Template Haskell code from GHC 8 to GHC 9 and analyzed their differences.
```
NaeosPsy:
```suggestion
In this post, we've seen the changes introduced in GHC 9 regarding Typed Template Haskell. We've learned about two different ways to migrate typed Template Haskell code from GHC 8 to GHC 9 and analyzed their differences.
```

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,205 @@
...
+
+Using `th-compat` has a big advantage in being compatible with two compilers, besides having an interface similar to the one in GHC 9. On the other hand, if supporting GHC 8 is not necessary, it becomes an extra dependency in your project, being potentially unfamiliar to many users. The choice of which strategy to use will ultimately depend on your specific needs, but we hope to have shed a light on their pros and cons.
```
NaeosPsy:
```suggestion
Using `th-compat` has a big advantage in being compatible with two compilers, besides having an interface similar to the one in GHC 9. On the other hand, if supporting GHC 8 is not necessary, it becomes an extra dependency in your project that's potentially unfamiliar to many users. The choice of which strategy to use will ultimately depend on your specific needs, but we hope to have shed light on their pros and cons.
```

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,205 @@
...
+
+For more Haskell tutorials, you can check out our [Haskell articles](https://serokell.io/blog/haskell) or follow us on [Twitter](https://twitter.com/serokell) or [Medium](https://serokell.medium.com/).
```
NaeosPsy:
```suggestion
For more Haskell tutorials, you can check out our [Haskell articles](https://serokell.io/blog/haskell) or follow us on [Twitter](https://twitter.com/serokell) or [Dev](https://dev.to/serokell/).
```

NaeosPsy (approved):

Introduction to Template Haskell/article.md
```diff
@@ -134,6 +134,8 @@ What is going on with the `$compose`? This is a _splice_, and it allows us to us
 
 Since Template Haskell is evaluated during compile time, if we replace `x` with a variable that was not bound, such as `z`, we get a `Variable not in scope: z` error instead during the splice generation. Any malformed expressions will be reported like any other regular Haskell expression. In other words, we analyze a Template Haskell declaration as if it was any other ordinary Haskell declaration.
 
+Keep in mind that in GHC 8, the usage of splices may require parentheses or not. For instance, if `compose` came from a qualified import, then you'd need to write `$(Foo.compose)` instead. In GHC 9, the parser is more relaxed and will not require parentheses in such situations.
```
kirelagin:
```suggestion
Keep in mind that before GHC 9, the usage of splices may require parentheses or not. For instance, if `compose` came from a qualified import, then you'd need to write `$(Foo.compose)` instead. In GHC 9, the parser is more relaxed and will not require parentheses in such situations.
```

GHC 7 (and even below) still exist.

Introduction to Template Haskell/article.md
```diff
@@ -327,6 +329,7 @@ generateTupleClass size = do
     r = mkName "r"
 
     -- class TupleX t r | t -> r where
+    -- In GHC 9: cDecl = ClassD [] className [PlainTV t (), PlainTV r ()] [FunDep [t] [r]] [mDecl]
```
kirelagin: Maybe move this line _below_ the next one (so that it does not visually stick to the other comment)? And also add a line break between the class declaration and the method declaration, again, just to separate them visually, now that there is this new GHC 9 comment.

Typed Template Haskell/article.md
```diff
@@ -68,7 +68,9 @@ SigE (LitE (IntegerL 42)) (ConT GHC.Base.String)
 
 ## Typed splices
 
-Just like we had untyped splices such as `$foo`, now we also have typed splices, written as `$$foo`. Note, however, that if your GHC version is below 9.0, you may need to write `$$(foo)` instead.
+Just like we had untyped splices such as `$foo`, now we also have typed splices, written as `$$foo`. In this section, we will see how to define and splice typed Template Haskell code.
+
+We reiterate one point mentioned in our previous article: in GHC 8, the usage of splices may require parentheses or not. For instance, if `foo` came from a qualified import, then you'd need to write `$$(Foo.foo)` instead. In GHC 9, the parser is more relaxed and will not require parentheses in such situations.
```
kirelagin:
```suggestion
We reiterate one point mentioned in our previous article: before GHC 9, the usage of splices may require parentheses or not. For instance, if `foo` came from a qualified import, then you'd need to write `$$(Foo.foo)` instead. In GHC 9, the parser is more relaxed and will not require parentheses in such situations.
```

Typed Template Haskell/article.md
```diff
@@ -295,6 +302,41 @@ Annotating `mempty'` may resolve it in this case:
 
 An [open ticket](https://gitlab.haskell.org/ghc/ghc/-/issues/10271) exists describing the issue, but if you run into some strange errors, it's a good idea to keep it in mind.
 
+## Compatibility with GHC 9
+
+In GHC 9, the types used by typed Template Haskell were changed according to the [Make Q (TExp a) into a newtype](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0195-code-texp.rst) proposal.
+
+This proposal changed the typed expression quasi-quoter so it now returns `Quote m => Code m a` instead of the old `Quote m => m (TExp a)` from [`template-haskell-2.17.0.0`](https://hackage.haskell.org/package/template-haskell-2.17.0.0/docs/Language-Haskell-TH.html#t:Code).
```
kirelagin: I don’t understand the link to th-2.17. I think, `Code` was implemented exactly at the same time as the overloaded bracket.

In other words, what I am saying is that the proposal you are referring indeed changed the type from `Quote m => m (TExp a)` to `Quote m => Code m a`, but also, a the exact same time another proposal changed the type from `Q (TExp a)` to `Quote m => m (TExp a)`, so the reference to 2.17 does not seem relevant, as both changes happened in there.

To sum up, IMHO, actionable points here are:

1. Maybe mention that it used to be `Q (TExp a)` prior to GHC 9.
2. Remove the link to template-haskell? I mean, I genuinely don’t understand what it means, maybe I am missing something.

Typed Template Haskell/article.md
```diff
@@ -295,6 +302,41 @@ Annotating `mempty'` may resolve it in this case:
...
+
+If the reader is curious, please take a moment to read our [Typed Template Haskell in GHC 9](https://serokell.io/blog/typed-template-haskell-in-ghc-9) article detailing a migration guide of typed TH code to GHC 9, where we discuss a backward-compatible way to integrate typed Template Haskell code in both GHC 8 and GHC 9.
```
kirelagin:
```suggestion
If the reader is curious, please take a moment to read our [Typed Template Haskell in GHC 9](./typed-template-haskell-in-ghc-9) article detailing a migration guide of typed TH code to GHC 9, where we discuss a backward-compatible way of integrating typed Template Haskell code in both GHC 8 and GHC 9.
```

I am not sure about absolute vs. relative URLs, what do we normally use? @NaeosPsy? I mean, I am personally a fan of relative URLs, I think we should use them whenever possible, but maybe there are reasons to prefer absolute?

kirelagin: > If the reader is curious, please take a moment to read

This is a weird transition from third person to second person imperative.

Typed Template Haskell/article.md
```diff
@@ -295,6 +302,41 @@ Annotating `mempty'` may resolve it in this case:
...
+... mempty' = [|| mempty ||]
+
+>>> x :: String
+... x = id $$mempty'
+
+>>> x
+""
+```
+
+Should you need to explicitly convert between the old and new types, you may pattern match directly on `Code` or use the following functions:
```
kirelagin:
```suggestion
Should you need to explicitly convert between the old and new types, you may pattern match directly on `Code` or use the following functions from [Language.Haskell.TH](https://hackage.haskell.org/package/template-haskell-2.18.0.0/docs/Language-Haskell-TH.html#t:Code):
```

kirelagin (commented):
Comments on the changes to the existing articles. Gonna read the new one now.

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,205 @@
+# Typed Template Haskell in GHC 9
+
+Welcome to our third post about Template Haskell! Today we will take a look at the changes that were made in GHC 9 regarding Typed Template Haskell (TTH) and how to use the [`th-compat`](https://hackage.haskell.org/package/th-compat) library to make TTH code that will work with both GHC 8 and GHC 9.
```
kirelagin:
```suggestion
Welcome to our third post about Template Haskell! Today we will take a look at the changes that were made in GHC 9 regarding Typed Template Haskell (TTH) and how to use the [`th-compat`](https://hackage.haskell.org/package/th-compat) library to write TTH code that will work with both GHC 8 and GHC 9.
```

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,205 @@
...
+
+In our [previous blog post](https://serokell.io/blog/typed-template-haskell-overview), we made an overview of Template Haskell in GHC 8. The article was later amended with the changes required so that the examples compile in GHC 9 in a non-backward compatible way. Make sure to read that post before you continue!
```
kirelagin:
```suggestion
In our [previous blog post](https://serokell.io/blog/typed-template-haskell-overview), we gave an overview of Typed Template Haskell in GHC 8. The article was later amended with the changes required so that the examples compile in GHC 9 in a non-backward compatible way. Make sure to read that post before you continue!
```

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,205 @@
...
+newtype Code m (a :: TYPE (r :: RuntimeRep)) = Code
+  { examineCode :: m (TExp a) -- ^ Underlying monadic value
+  }
+```
+
+The differences between the simplified and real definitions are not important for the understanding of this post, and you may keep the simplified definition in mind while reading through this.
+
+Thankfully, the untyped TH interface remains mostly unchanged, and the typed TH interface should only need to use `Code` instead of `TExp` to compile again with some helper functions.
+
+<!-- TODO: Write about motivation as well? -->
```
kirelagin: No, I think it’s fine, the motivation is in the proposal that you link.

Introduction to Template Haskell/article.md
heitor-lassarote: Done, I think it's better now.

Typed Template Haskell/article.md
heitor-lassarote: I've removed the link and used `Q (TExp a)` there.

Typed Template Haskell/article.md
heitor-lassarote: Changed "the reader is" to "you're".

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,205 @@
...
+    Nothing -> [|| Production ||]  -- We default to Production if the flag is unset
+    Just level -> case level of
+      "DEVELOPMENT" -> [|| Development ||]
+      "PRODUCTION" -> [|| Production ||]
+      other -> fail $ "Unrecognized LOG_LEVEL flag: " <> other
+```
+
+If you try to compile this module with GHC 9, however, this example will fail to compile:
+
+```hs
```
kirelagin: This might or might not be a good idea. I would guess that `text` would be a safer bet.

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,205 @@
...
+    • In the Template Haskell quotation [|| Production ||]
+      In the expression: [|| Production ||]
+      In a case alternative: "PRODUCTION" -> [|| Production ||]
+   |
+21 |       "PRODUCTION" -> [|| Production ||]
+   |                       ^^^^^^^^^^^^^^^^^^
+Failed, no modules loaded.
+```
+
+In the next sections, we will see how to port this code to GHC 9 without and with backward compatibility with GHC 8.
```
kirelagin:
```suggestion
In the following sections, we will see how to port this code to GHC 9 without and with backward compatibility with GHC 8.
```

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,205 @@
...
+```hs
+import Language.Haskell.TH.Syntax (Code, Q, examineCode, liftCode)
+```
+
+Since we now need to return a value of type `Code`, we change our type signature. And now we can fix the code simply by using two new functions, namely `examineCode` and `liftCode`:
+
+```diff
+-logLevelFromFlag :: Q (TExp LogLevel)
+-logLevelFromFlag = do
++logLevelFromFlag :: Code Q LogLevel
```
kirelagin, line 111: Ok, so, I think that the article thus far has not justified this change. Namely, it only ever says that _quotations_ changed their types, so the reader might be wondering why we can’t stay in the `Q` monad (and the answer is that we can, but it won’t be very useful, since _splices_ now want `Code` too).

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,205 @@
...
+```
+
+In the next sections, we will see how to port this code to GHC 9 without and with backward compatibility with GHC 8.
+
+## Porting old TTH code without backward compatibility
+
+For this section, you need a GHC whose version is at least 9. Make sure to import the appropriate definitions from `template-haskell` as well:
+
+```hs
+import Language.Haskell.TH.Syntax (Code, Q, examineCode, liftCode)
```
kirelagin: Maybe make this a diff too?

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,205 @@
...
++logLevelFromFlag :: Code Q LogLevel
++logLevelFromFlag = liftCode $ do  -- liftCode will turn 'Q' into 'Code'
+  flag <- liftIO $ lookupEnv "LOG_LEVEL"
+  case flag of
++    -- examineCode will turn 'Code' into 'Q'
+-    Nothing -> [|| Production ||]  -- We default to Production if the flag is unset
++    Nothing -> examineCode [|| Production ||]  -- We default to Production if the flag is unset
+    Just level -> case level of
+_      "DEVELOPMENT" -> [|| Development ||]
+_      "PRODUCTION" -> [|| Production ||]
```
kirelagin:
```suggestion
-      "DEVELOPMENT" -> [|| Development ||]
-      "PRODUCTION" -> [|| Production ||]
```

😮

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,205 @@
...
++      "DEVELOPMENT" -> examineCode [|| Development ||]
++      "PRODUCTION" -> examineCode [|| Production ||]
+      other -> fail $ "Unrecognized LOG_LEVEL flag: " <> other
+```
+
+The errors in the previous section happened because the type returned by the quotations is no longer a `Q`, but `Code` instead. The solution is to use `examineCode` to turn a `Code m a` into a `m (TExp a)`, and then `liftCode` to do the opposite operation: turning a `m (TExp a)` into a `Code m a`.
```
kirelagin: Maybe instantiate `m` with `Q` everywhere here, since that’s what you have in your code.

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,205 @@
...
+```hs
+liftCode    :: forall (r :: RuntimeRep) (a :: TYPE r) m. m (TExp a) -> Code m a
+examineCode :: forall (r :: RuntimeRep) (a :: TYPE r) m. Code m a -> m (TExp a)
+```
+
+And that's it! If you want the code to still work on GHC 8, however, make sure to read the next section.
+
+## Porting old TTH code with backward compatibility
+
+For this section, you may choose to use either GHC 8 or GHC 9. We will use the [`th-compat`](https://hackage.haskell.org/package/th-compat) package, besides the familiar [`template-haskell`](https://hackage.haskell.org/package/template-haskell) package, so make sure you load them.
```
kirelagin: Maybe explain that this package is not some random package from the internets, but the “official” migration path from the proposal.

kirelagin:
```suggestion
For this section, you may choose to use either GHC 8 or GHC 9. We will use the [`th-compat`](https://hackage.haskell.org/package/th-compat) package, besides the familiar [`template-haskell`](https://hackage.haskell.org/package/template-haskell) package, so make sure you load them both.
```

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,205 @@
...
+
+The pattern is pretty similar to the workflow with `Code`, albeit with `Splice` now.
```
kirelagin, line 141: Kinda feels like some intro is missing here. Like, just a couple of words about this package providing `Splice`, which is magically defined in such a way that it works on either version of GHC.

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,205 @@
...
++    Nothing -> examineSplice [|| Production ||]  -- We default to Production if the flag is unset
+    Just level -> case level of
+-      "DEVELOPMENT" -> examineCode [|| Development ||]
+-      "PRODUCTION" -> examineCode [|| Production ||]
++      "DEVELOPMENT" -> examineSplice [|| Development ||]
++      "PRODUCTION" -> examineSplice [|| Production ||]
+      other -> fail $ "Unrecognized LOG_LEVEL flag: " <> other
+```
+
+And that's it! `Splice m a` is defined in `th-compat` as `m (TExp a)` in GHC 8 and `Code m a` in GHC 9, and this is why this code works. In addition, functions like `liftSplice` and `examineSplice` will either be defined `liftCode` and `examineCode` (in GHC 9) or `id` (in GHC 8).
```
kirelagin: My usual nitpick: it is not about GHC 9 vs. GHC 8 – it is about GHC >=9 vs. GHC < 9.

kirelagin:
```suggestion
And that's it! `Splice m a` is defined in `th-compat` as `m (TExp a)` in GHC 8 and `Code m a` in GHC 9, and this is why this code works. In addition, functions like `liftSplice` and `examineSplice` will either be defined as `liftCode` and `examineCode` (in GHC 9) or `id` (in GHC 8).
```

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,205 @@
...
+
+<interactive>:200:1: error: parse error on input ‘$$’
+```
+
+But it works in GHC 9:
+
+```hs
+>>> $$TH.logLevelFromFlag
+Production
+```
```
kirelagin, line 191: Again, feels like some explanation is missing. Like, just say what you said in other articles, that this is because they tweaked the parser.

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Conclusion
+
+In this post, we've seen the changes introduced in GHC 9 regarding Typed Template Haskell. We've learned about two different ways to migrate typed Template Haskell code from GHC 8 to GHC 9 and analyzed their differences.
+
+Using `th-compat` has a big advantage in being compatible with two compilers, besides having an interface similar to the one in GHC 9. On the other hand, if supporting GHC 8 is not necessary, it becomes an extra dependency in your project that's potentially unfamiliar to many users. The choice of which strategy to use will ultimately depend on your specific needs, but we hope to have shed light on their pros and cons.
```
kirelagin: > with two compilers

meh

kirelagin (commented):
Apart from a couple of minor comments, the content looks good to me.

I remember that I asked for an article comparing two approaches to compatibility, but they were not “without compatibility” and “with compatibility” 🙂. They were “using `IsCode`” and “using `Splice`. I am still not sure why this entire `IsCode` story exists, since `Splice` seems to cover most (if not all) use-cases and, more importantly, libraries providing TH expressions usually need to provide something spliceable in the end. So, I think, what we already have is good and is enough, however personally I am still looking for an answer to my question: what is the situation in which I would not want to use `Splice` and would want to use `toCode`/`fromCode` instead.

typed-th-in-ghc-9/article.md
kirelagin: Maybe even, in addition to the function definition, show an example splicing it, and the demonstrate that with the original code on GHC9 you are not only getting type mismatches from quotations, but from splices too.

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,205 @@
...
+    • Couldn't match type ‘Code m2’ with ‘Q’
+      Expected: Q (TExp LogLevel)
+        Actual: Code m2 LogLevel
+    • In the Template Haskell quotation [|| Production ||]
+      In the expression: [|| Production ||]
+      In a case alternative: "PRODUCTION" -> [|| Production ||]
+   |
+21 |       "PRODUCTION" -> [|| Production ||]
+   |                       ^^^^^^^^^^^^^^^^^^
+Failed, no modules loaded.
```
kirelagin: What I do not particularly like about this example is that it results in three errors that are exactly the same. Can’t we come up with something that will just give one error? I think one error is convincing enough and we don’t need to waste space on three errors.

typed-th-in-ghc-9/article.md
heitor-lassarote: Oh, what a facepalm... thanks for suggesting this change.

Typed Template Haskell/article.md
kirelagin: Ok, but now the connection between the old time and the new type is harder to grasp... at least without any context. Maybe it would be worth it to somehow point out that `m` is what used to be `Q` before the changes?

heitor-lassarote: I elaborated a bit more. I think it should be clearer now.

typed-th-in-ghc-9/article.md
heitor-lassarote: Got it. Removed this TODO.

typed-th-in-ghc-9/article.md
heitor-lassarote: For a minute, I was pretty confused thinking that you meant the `text` package and I couldn't understand where it should be used. Now I see you meant using `text` syntax instead of `hs`. I've changed it.

typed-th-in-ghc-9/article.md
heitor-lassarote: Good idea, I changed this to a diff, and in the imports in the next section as well.

typed-th-in-ghc-9/article.md
heitor-lassarote: Done. I kept the `liftCode` and `examineCode` snippets below unchanged to reflect their actual polymorphic types.

typed-th-in-ghc-9/article.md
heitor-lassarote: Could you elaborate on this point? That is, I'm not sure how to indicate that this is the "official" migration path, so I wrote that we (Serokell) recommend it, for now. This should also hopefully resolve your suggestion regarding adding an intro to this section.

typed-th-in-ghc-9/article.md
heitor-lassarote: Changed it.

typed-th-in-ghc-9/article.md
heitor-lassarote: I changed the phrasing a bit to indicate it was tweaked, I hope it's clearer now.

typed-th-in-ghc-9/article.md
heitor-lassarote: I changed to GHC versions, I hope it's better.

typed-th-in-ghc-9/article.md
heitor-lassarote: I've trimmed the other two errors and added a comment mentioning they are omitted.

typed-th-in-ghc-9/article.md
heitor-lassarote: I added a simple example, I hope a hole is a good way to demonstrate that `Code` is what is now expected.

heitor-lassarote:
@kirelagin, I'm looking a bit more about `IsCode`, although I also don't completely understand it. My guess is that `th-compat` defined `IsCode` and likely kept it there for compatibility reasons, as `Splice` didn't originally exist, and perhaps to overcome some of its limitations (described in its documentation). I see one existed but not the other since the first-ever pull request: https://github.com/haskell-compat/th-compat/pull/1/files

`Splice` is used in the package's `README` and in most tests, so I'd guess this is what should be used as there is no mention of `IsCode` in the `README`, but I could be wrong. My only worry is in `Splice`'s Haddock which says:

> This should be used with caution, as its definition differs depending on which version of `template-haskell` you are using. It is mostly useful for contexts in which one is writing a definition that is intended to be used directly in a typed Template Haskell splice, as the types of TH splices differ between `template-haskell` versions as well.

There is also this when `Splice` was initially introduced:

> These are primarily useful for defining liftTyped manually without needing excessive amounts of CPP.

Although I'm not sure if this is the motivation for the introduction of `Splice` or `liftTypedFromUntypedSplice` which came from the same PR.

I don't see the problem if using the appropriate functions as in this article and in the `README`. If I understand correctly, if you don't intend to directly splice it, then maybe `Code` should be used? Please correct me if I am wrong. From the docs about `Splice`:

> This section of code is useful for library authors looking to provide a typed TemplateHaskell interface that is backwards- and forward-compatible. This section may be useful for you if you specifically intend for the splice to be done directly.

Typed Template Haskell/article.md
NaeosPsy: We have historically used absolute (read: I have never imagined to use a relative link like that, nor has anyone else before, I think), and I think it would be simpler to leave it at status quo.

kirelagin: Ok, but if we rename the company and change the domain, you’ll have to go fix it all 🙄

kirelagin (approved):

typed-th-in-ghc-9/article.md
```diff
@@ -0,0 +1,199 @@
...
+
+And that's it! `Splice m a` is defined in `th-compat` as `Code m a` in GHC 9 and as `m (TExp a)` in prior versions, and this is why this code works. In addition, functions like `liftSplice` and `examineSplice` will either be defined as `liftCode` and `examineCode` (in GHC 9) or `id` (in GHC 8 and below).
+
+## Parentheses in splices
+
+Just one more thing before we conclude this article. You may see surprising behavior regarding the usage of parentheses in splices for both untyped and typed Template Haskell. The parser was tweaked in GHC 9 so parentheses are unnecessary in some situations compared to prior versions.
+
+For example, if we had imported `logLevelFromFlag` qualified, then you'd write `$$(TH.logLevelFromFlag)` in GHC 8, otherwise you'd get a parser error:
+
+```hs
```
kirelagin: Unhighlight this code block and the one below.

kirelagin: cc @NaeosPsy

PR #8, opened 2021-12-15
[PR-1246] Article about Free Monads

lierdakil:
~~I still need to go over this once or twice before I'm feeling ready to show it to anyone, but I thought I should at least put this on the table, so to speak.~~

The first draft is ready for review.

### DIFF
--- a/Introduction to Free Monads/article.md
+++ b/Introduction to Free Monads/article.md
@@ -46,7 +46,7 @@ Second, to in some sense combine two elements, we have `(<>) :: α -> α -> α`.
 
 <p>
 <details>
-<summary>A note about (&lt;&gt;)</summary>
+	<summary>A note about <code>(<>)</code></summary>
 
 The `(<>)` operator is actually defined in the `Semigroup` type class, but we'll gloss over semigroups here for the sake of simplicity.
 
@@ -73,7 +73,7 @@ Indeed, if the set $M$ is a set of lists of elements of $S$:
 type M = [S]
 ```
 
-then our embedding can be defined as:
+Then our embedding can be defined as:
 
 ```haskell
 i :: S -> M
@@ -142,14 +142,14 @@ m >>= return = m -- right identity
 If you look at the monad laws and squint a bit (or perhaps a lot), you might notice that these laws seem somewhat similar to the monoid laws.
 The names of the laws are the same, at least.
 Our choice of the definition makes it a little harder to see the connection, though.
-If we instead rewrite these laws using the Kleisli composition
+If we instead rewrite these laws using the Kleisli composition:
 
 ```haskell
 (>=>) :: Monad m => (a -> m b) -> (b -> m c) -> (a -> m c)
 f >=> g = \x -> f x >>= g
 ```
 
-it becomes much more apparent:
+It becomes much more apparent:
 
 ```haskell
 (f >=> g) >=> h = f >=> (g >=> h) -- associativity
@@ -205,7 +205,7 @@ data List a = Nil | Cons a (List a)
 ```
 
 Now, `List a` is a type, but `Free f` is a unary type constructor, so we'd need one more type argument.
-Doing that directly yields
+Doing that directly yields:
 
 ```haskell
 data Free f a = Nil a | Cons f (Free f) a
@@ -453,7 +453,7 @@ printFreeList (Free f) = "["
   <> "]"
 ```
 
-if we run `printFreeList listComputation`, we will get
+If we run `printFreeList listComputation`, we will get:
 
 ```haskell
 [[[111,211],[121,221]],[[112,212],[122,222]],[[113,213],[123,223]]]
@@ -957,4 +957,4 @@ All examples from this article are available as code [on GitHub][github-link].
 
     Implement a conventional `Monad` instance, and implement `convert :: FreeBinTree l a -> BinTree l a` using `foldFree` and a natural transformation `BinTreeF l a -> BinTree l a`.
 
-If you get absolutely stuck, the intended solutions to exercises are avaliable in the [GitHub repository with the examples][github-link] on the `solutions` branch.
+If you get absolutely stuck, the intended solutions to exercises are avaliable in the [GitHub repository with the examples][github-link] in the `solutions` branch.
--- a/Introduction to Free Monads/article.md
+++ b/Introduction to Free Monads/article.md
@@ -300,6 +300,7 @@ liftF :: (Functor f, MonadFree f m) => f a -> m a
 ```
 
 `MonadFree` is just an MTL-style type class for different encodings of the free monad.
+Also, while on the topic of different encodings, for more flexibility, instead of directly using `Free` and `Pure`, one would generally use `wrap` (defined in `MonadFree`) and `pure` respectively.
 In our case, we can pretend that:
 
 ```haskell
@@ -580,6 +581,7 @@ if' cond t f = Free $ If cond t f
 ```
 
 Notice how instead of `liftF` we have to use `Free` directly here -- our branches are already of type `FreeAST`.
+As discussed earlier, in general you would use `wrap` defined in `MonadFree` instead of `Free` directly, but we're glossing over that for simplicity.
 We can now modify our interpreter `computeAST` by adding another case:
 
 ```haskell
--- a/Introduction to Free Monads/article.md
+++ b/Introduction to Free Monads/article.md
@@ -889,4 +889,4 @@ See [Free and Freer Monads: Putting Monads Back into Closet
 
     It is a monad.
 
-    Implement a conventional `Monad` instance, and implement `convert :: FreeBinTree l a -> BinTree l a` using `foldFree` and a natural transformation.
+    Implement a conventional `Monad` instance, and implement `convert :: FreeBinTree l a -> BinTree l a` using `foldFree` and a natural transformation `BinTreeF l a -> BinTree l a`
--- a/Introduction to Free Monads/article.md
+++ b/Introduction to Free Monads/article.md
@@ -26,27 +26,34 @@ This part is a bit theory-heavy, so if it is not your cup of tea, feel free to s
 
 In abstract algebra, we call something a "free X" when it is in some sense a minimal structure that satisfies conditions for being X.
 The exact definition of "minimal" in abstract algebra is a little vague, but the main idea is that "free X" satisfies all laws for X and does not have any other laws describing it.
-That is, in this context, "free" means "unrestricted" -- it is not that we can get a structure "for free"[^for-free], but that no other laws are restricting the structure apart from those absolutely necessary.
-
-[^for-free]: Although in most cases, we actually might, depending on your definition of "for free".
+That is, in this context, "free" means "unrestricted" -- it is not that we can get a structure "for free", but that no other laws are restricting the structure apart from those absolutely necessary.
+That said, depending on your definition of "for free", in most cases, we actually might also get a structure "for free".
 
 More formally, a free structure over a set $S$ is a set $M,$ together with operations on elements of $M,$ such that:
 
-- There is an embedding[^embedding] $i: S\to M$.
+- There is an embedding $i: S\to M$.
 - $M$ is a minimal closure, i.e., it only contains all the results of applying $i$ to elements of $S$, and any results of applying the operations, defined as part of the structure, to the elements of $M$.
 - The only laws that hold for the generated structure are those required to hold.
 
-[^embedding]: An embedding is an injective morphism.
-    That is, an embedding $S\to M$ is a mapping from elements of $S$ to elements of $M$, such that each element of $S$ is mapped onto a unique element of $M$ (injectivity), and, in some sense, this mapping is structure-preserving (i.e. it's a morphism).
-    The point is that every element of the original set is uniquely representable in the free structure.
+An embedding is an injective morphism.
+That is, an embedding $S\to M$ is a mapping from elements of $S$ to elements of $M$, such that each element of $S$ is mapped onto a unique element of $M$ (injectivity), and, in some sense, this mapping is structure-preserving (i.e. it's a morphism).
+The point is that every element of the original set is uniquely representable in the free structure.
 
 Let us start with a simple example, specifically, `Monoid`s.
 In Haskell, any type `α` in the `Monoid` type class must have two functions defined.
 First, to construct a neutral element, we have `mempty :: α`.
-Second, to in some sense combine two elements, we have `(<>) :: α -> α -> α`[^semigroup].
-Additionally, these functions must satisfy three equations, called the monoid laws:
+Second, to in some sense combine two elements, we have `(<>) :: α -> α -> α`.
 
-[^semigroup]: The `(<>)` operator is actually defined in the `Semigroup` type class, but we'll gloss over semigroups here for the sake of simplicity.
+<details>
+<summary>A note about `(<>)`</summary>
+
+The `(<>)` operator is actually defined in the `Semigroup` type class, but we'll gloss over semigroups here for the sake of simplicity.
+
+***
+
+</details>
+
+Additionally, these functions must satisfy three equations, called the monoid laws:
 
 ```haskell
 (x <> y) <> z = x <> (y <> z) -- associativity
@@ -103,15 +110,22 @@ If you're interested, [Bartosz Milewski's article on adjunctions][adjunctions] d
 Before we see how this idea applies to monads, it might be helpful to compare monads to monoids first.
 
 Let us first review the definition of the `Monad` type class.
-In Haskell, it is defined as[^applicative]:
-
-[^applicative]: If you're more familiar with modern Haskell, you might object that actually, the superclass of `Monad` is `Applicative`, but we will gloss over that for the sake of simplicity.
+In Haskell, it is defined as:
 
 ```haskell
 class Functor m => Monad m where
   return :: a -> m a
   (>>=) :: m a -> (a -> m b) -> m b
 ```
+
+<details>
+<summary>A note on `Monad`'s superclass</summary>
+
+If you're more familiar with modern Haskell, you might object that actually, the superclass of `Monad` is `Applicative`, but we will gloss over that for the sake of simplicity.
+
+***
+
+</details>
 
 This definition needs to additionally satisfy the monad laws:
 
@@ -131,12 +145,7 @@ If we instead rewrite these laws using the Kleisli composition
 f >=> g = \x -> f x >>= g
 ```
 
-it becomes much more apparent[^equivalence]:
-
-[^equivalence]: The proof of equivalence is left as an exercise to the reader.
-    Here, by equivalence we mean that the monad laws expressed via Kleisli arrows imply the monad laws expressed via `>>=` and vice versa.
-    The proof itself is a straightforward application of equational reasoning.
-    It's easier to start from the Kleisli arrow form.
+it becomes much more apparent:
 
 ```haskell
 (f >=> g) >=> h = f >=> (g >=> h) -- associativity
@@ -144,14 +153,34 @@ return >=> f = f -- left identity
 f >=> return = f -- right identity
 ```
 
+(the proof of equivalence between monad laws expressed via `>>=` and `>=>` is left as an exercise for the reader)
+
+<details>
+<summary>On equivalence</summary>
+
+Here, by equivalence we mean that the monad laws expressed via Kleisli arrows imply the monad laws expressed via `>>=` and vice versa.
+The proof itself is a straightforward application of equational reasoning.
+It's easier to start from the Kleisli arrow form.
+
+***
+
+</details>
+
 That is why a monad is indeed a monoid: it satisfies the same laws, and the "values" are of the type `Functor f => a -> f b`.
 Note, though, that this is not the category of endofunctors.
-However, there exists a natural transformation to the category of endofunctors[^endofunctors].
+However, there exists an isomorphism with the category of endofunctors.
 
-[^endofunctors]: One could alternatively define monads via `join :: m (m a) -> m a` instead of `>>=`, which would make it a little more evident that monads are monoids over endofunctors, but to explain this properly, we would need to use a bit more of category theory jargon.
-    There is an excellent [r/math post][endofunctors] on the subject.
+<details>
+<summary>On endofunctors</summary>
 
-    [endofunctors]: https://www.reddit.com/r/math/comments/ap25mr/a_monad_is_a_monoid_in_the_category_of/
+One could alternatively define monads via `join :: m (m a) -> m a` instead of `>>=`, which would make it a little more evident that monads are monoids over endofunctors, but to explain this properly, we would need to use a bit more of category theory jargon.
+There is an excellent [r/math post][endofunctors] on the subject.
+
+[endofunctors]: https://www.reddit.com/r/math/comments/ap25mr/a_monad_is_a_monoid_in_the_category_of/
+
+***
+
+</details>
 
 We can guess a few things based on the intuition we gained by looking at free monoids.
 First, we might note that `return` and `>=>` correspond to `mempty` and `(<>)`, respectively.
@@ -380,15 +409,23 @@ There's no free lunch here, though, so we need to define the semantics of this m
 ### List as a free monad
 
 Not all free monads behave exactly the same as their regular counterparts.
-For instance, you might know that the list monad encodes non-determinism[^non-determinism].
+For instance, you might know that the list monad encodes non-determinism.
+
+<details>
+<summary>In case the last statement is surprising...</summary>
+
+If the statement about list monad encoding non-determinism is surprising, consider that one could use a list to represent all possible outcomes of some event and then handle those one by one.
+For further reading on this topic, refer to the [School of Haskell article on the list monad][school-of-haskell] or the [Haskell wikibook section on understanding monads][haskell-wikibook]
+
+[school-of-haskell]: https://www.schoolofhaskell.com/school/starting-with-haskell/basics-of-haskell/13-the-list-monad#non-deterministic-computations
+[haskell-wikibook]: https://en.wikibooks.org/wiki/Haskell/Understanding_monads/List
+
+***
+
+</details>
+
 However, if we derive a free monad for lists, the behavior is slightly different.
 To give a specific example:
-
-[^non-determinism]: If the statement about list monad encoding non-determinism is surprising, consider that one could use a list to represent all possible outcomes of some event and then handle those one by one.
-    For further reading on this topic, refer to the [School of Haskell article on the list monad][school-of-haskell] or the [Haskell wikibook section on understanding monads][haskell-wikibook]
-
-    [school-of-haskell]: https://www.schoolofhaskell.com/school/starting-with-haskell/basics-of-haskell/13-the-list-monad#non-deterministic-computations
-    [haskell-wikibook]: https://en.wikibooks.org/wiki/Haskell/Understanding_monads/List
 
 ```haskell
 listComputation :: Free [] Int
@@ -422,18 +459,27 @@ In general, we can restore any proper monad behavior from a free monad since it 
 foldFree :: Monad m => (forall x. f x -> m x) -> Free f a -> m a
 ```
 
-Given a function[^natural-transformation] converting `f x` into some monad `m x` for any `x`, we can convert `Free f a` into `m a`.
+Given a function (a natural transformation) converting `f x` into some monad `m x` for any `x`, we can convert `Free f a` into `m a`.
+
+<details>
+<summary>On natural transformations</summary>
+
+The discussion of naturality is not particularly relevant here, but we'll introduce the concept briefly.
+The natural transformation is a structure-preserving transformation between two functors.
+Here, the naturality condition is enforced by the parametricity, that is, the function is polymorphic in `x`.
+
+Essentially, you can think of any function with type `forall a. f a -> g a` as a natural transformation between `f` and `g`, as long as you ignore `undefined` and other bottoms.
+
+***
+
+</details>
+
 Returning to our list example, we can get the original list behavior by applying `foldFree id`:
 
 ```haskell
 > foldFree id listComputation
 [111,211,121,221,112,212,122,222,113,213,123,223]
 ```
-
-[^natural-transformation]: This function is a natural transformation.
-    The naturality condition is enforced by parametricity.
-    That being said, the discussion of naturality is not particularly relevant here.
-    Essentially, you can think of any function with type `forall a. f a -> g a` as a natural transformation between `f` and `g`, as long as you ignore `undefined` and other bottoms.
 
 ### Free monads for EDSLs
 
@@ -464,10 +510,8 @@ If we try to derive a `Functor` instance, GHC will complain:
 Constructor ‘Add’ must not use the type variable in a function argument
 ```
 
-It is, in fact, impossible to define a lawful `Functor` instance if the functor's type variable ends up in the function argument position[^variance].
+It is, in fact, impossible to define a lawful `Functor` instance if the functor's type variable ends up in the function argument position (this has to do with `Functor` being a covariant functor, but discussion of variance is out of scope here).
 Hence, we will introduce two type variables: one for operation arguments, and another for the overall result.
-
-[^variance]: This has to do with `Functor` being a covariant functor, but discussion of variance is out of scope here.
 
 ```haskell
 data ASTF t a
--- a/Introduction to Free Monads/article.md
+++ b/Introduction to Free Monads/article.md
@@ -214,7 +214,8 @@ data Free f a = Pure a | Free (f (Free f a))
 
 As you can see, it is indeed very similar to a list.
 Note, however, that it is a bit more general: the functor is arbitrary, potentially making it a tree with branches of type `f` and leaves of type `a`.
-Pure values, i.e. leaves, are encoded via the constructor `Pure`, and "actions", i.e. branches -- via the constructor `Free`.
+Leaves are encoded via the `Pure` constructor, and branches -- via the `Free` constructor.
+We can also note that leaves correspond to pure values, and branches correspond to monadic "actions", respectively.
 
 This point is important, so we'll repeat: free monad is similar to a list, but, unlike a list, "continuation" (i.e. `Free`) can be a branching structure, depending on the choice of the base functor `f`.
 
--- a/Introduction to Free Monads/article.md
+++ b/Introduction to Free Monads/article.md
@@ -525,7 +525,7 @@ computeAST = foldFree go
         pure next
 ```
 
-We can write other interpreters, of course. 
+We can write other interpreters, of course.
 For instance, we could write a pretty-printer here.
 Using the fact that we can convert a free monad to any other monad using `foldFree`, we'll do just that.
 For instance, let's convert `FreeAST String` into `WriterT String (State Int)`.
@@ -667,7 +667,7 @@ output x1
 
 The reason for this is simple: when the free monad is built up, the continuation is passed to each place where the base functor is recursive in its parameter.
 This means that all the code after our "if'" command gets copied to both branches.
-There isn't a workaround for this because free monads build trees and not general graphs. 
+There isn't a workaround for this because free monads build trees and not general graphs.
 However, a more efficient representation -- e.g., the Church encoding mentioned previously -- will reduce the overhead.
 
 #### Decomposing ASTs
@@ -828,6 +828,10 @@ This construction is known as freer monads (as in, more free than free).
 See [Free and Freer Monads: Putting Monads Back into Closet
 ](https://okmij.org/ftp/Computation/free-monad.html).
 
+All examples from this article are available as code [on GitHub][github-link].
+
+[github-link]: https://github.com/lierdakil/free-monad-examples
+
 ## Exercises
 
 1. Implement the standard monads `Reader` and `Writer` using `Free`.
@@ -892,3 +896,5 @@ See [Free and Freer Monads: Putting Monads Back into Closet
     It is a monad.
 
     Implement a conventional `Monad` instance, and implement `convert :: FreeBinTree l a -> BinTree l a` using `foldFree` and a natural transformation `BinTreeF l a -> BinTree l a`.
+
+If you get absolutely stuck, the intended solutions to exercises are avaliable in the [GitHub repository with the examples][github-link] on the `solutions` branch.
--- a/Introduction to Free Monads/article.md
+++ b/Introduction to Free Monads/article.md
@@ -378,9 +378,15 @@ There's no free lunch here though, so we need to define the semantics of this mo
 ### List as a free monad
 
 Not all free monads behave exactly the same as their regular counterparts.
-For instance, you might know that the list monad encodes non-determinism.
+For instance, you might know that the list monad encodes non-determinism[^non-determinism].
 However, if we derive a free monad for lists, the behavior is slightly different.
 Consider:
+
+[^non-determinism]: If the statement about list monad encoding non-determinism is surprising, consider that one could use a list to represent all possible outcomes of some event, and then handle those one by one.
+    For further reading on this topic, refer to the [School of Haskell article on the list monad][school-of-haskell] or to the [Haskell wikibook section on understanding monads][haskell-wikibook]
+
+    [school-of-haskell]: https://www.schoolofhaskell.com/school/starting-with-haskell/basics-of-haskell/13-the-list-monad#non-deterministic-computations
+    [haskell-wikibook]: https://en.wikibooks.org/wiki/Haskell/Understanding_monads/List
 
 ```haskell
 listComputation :: Free [] Int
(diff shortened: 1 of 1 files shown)


Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,702 @@
+# Introduction to Free Monads
+
+If you've been around Haskell circles for a bit, you've probably seen the term "free monads".
```
NaeosPsy, line 3: This is rather meandering, perhaps just list some of the purposes Free Monads can be used for (as opposed to more common types of monads)?

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,702 @@
...
+[hackage-link]: https://hackage.haskell.org/package/free
+[stackage-link]: https://www.stackage.org/lts/package/free
+
+Using free monads, we can define a computation as a data structure.
+Here, "computation" is a very broad term.
+The data structure in question doesn't define how the computation is performed, and we can write a separate interpreter (or indeed many interpreters) that performs the actual computation in question.
+
+## Using free monads
+
+> Talk is cheap, show me the code
```
NaeosPsy: My gut tells me to cut this quote, it is rather overquoted & in comparison to others doesn't really bring anything new.

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,702 @@
...
+
+[^natural-transformation]: This function is a natural transformation.
+    The naturality condition is enforced by parametricity.
+    That being said, the discussion of naturality is not particularly relevant here.
+    Essentially, you can think of any function with type `forall a. f a -> g a` as a natural transformation between `f` and `g`, as long as you ignore `undefined` and other bottoms.
+
+### Free monads for EDSLs
+
+By now, we hopefully got some intuition for how free monads work.
+Let us now define a toy calculator language, which reads a few integers from the standard input, adds them together, and prints the result.
```
NaeosPsy:
```suggestion
Let us now define a toy calculator language that reads a few integers from the standard input, adds them together, and prints the result.
```

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,702 @@
...
+        Nil))
+  (8 (7 6
+        Nil)
+     (10 9
+         Nil))
+```
+
+## Conclusions
+
+> The only way to learn a new programming language is by writing programs in it.
```
NaeosPsy, line 868: Similarly, this quote can be cut as well, but I am not so much opposed to it as the Linus' one.

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,702 @@
...
+>
+> -- Dennis Ritchie
+
+Hopefully, after this brief introduction, you understand, at least in principle, what free monads are, and how they may be useful.
+Of course, the only way to get proficient with them is to use them in projects, and no blog post can replace actual hands-on experience, but that applies to everything in programming.
```
NaeosPsy, line 873: Are there some exercises that you could add? We are big on exercises at Serokell Blog. :D

NaeosPsy:
Added a few conceptual comments, grammar edit will come after developer reviews.

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,702 @@
...
+
+In abstract algebra, we call something a "free X" when it is in some sense a minimal structure that satisfies conditions for being X.
+The exact definition of "minimal" in abstract algebra is a little vague, but the main idea is that "free X" satisfies all laws for X, and does not have any other laws describing it.
+That is, in this context, "free" means "unrestricted" -- it is not that we can get a structure "for free"[^for-free], but that no other laws are restricting the structure apart from those absolutely necessary.
+
+[^for-free]: Although in most cases, we actually might, depending on your definition of "for free".
+
+More formally, a free structure over a set $S$ is a set $M,$ together with operations on elements of $M,$ such that:
+
+- There is an embedding $i: S\to M$.
```
alyoanton9, line 34: For me, as a novice in abstract algebra, "embedding" wasn't clear from the first glance. I got it's some mapping, but then went to google it. So, it would be nice to have a relevant link here.

Moreover, is it really `S -> M`, not `M -> S`?

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,702 @@
...
+
+It may be helpful to write out the types of these constructors explicitly:
+
+```haskell
+Pure :: a -> Free f a
+Free :: f (Free f a) -> Free f a
+```
+
+If you squint a bit, you will notice that `Pure` has the same signature as `return` (assuming `Monad (Free f)`).
+If you squint some more, you will notice that the signature of `Free` is very similar to the signature of `join`:
```
alyoanton9, line 238: Roughly speaking, `Cons` concatenates the single element and a list of elements, while `join` combines many elements (list of elements for the simplicity). Comparing `Cons` with `Free` looks quite natural to me, as it encodes action followed by the sequence of actions, so that comparing `Free` later with `join` confuses me. I see, that it's about similar signatures, but still feel a bit strange about it. Consequently, I don't get any insight from this analogue.

> this hopefully provides some insight into why such a structure works as an encoding for a free monad

Perhaps, not mentioning `join` would look better __for me__.  Anyway, this is my personal feeling.

Introduction to Free Monads/article.md
lierdakil: Embedding is an injective morphism, I can just add a footnote explaining that. The point is every element in the original set is uniquely representable in the free structure, so it has to be `S → M` -- were it `M → S` there is no guarantee that we could uniquely (or at all) represent elements of the original set in the structure, which would make the claim that it's the structure over set `S` highly dubious.

IIRC below I give the example of a free monoid over type `T` being a list in Hask, where the embedding is `embed x = [x]` (i.e. essentially `pure`) -- the function isn't bijective mind, there's no injective inverse `unembed :: [a] -> a`, so that hopefully illustrates the point of embedding having to be `S → M` and not vice versa (here `S` being `a` ∀ a ∈ Hask and `M` being `[a]` ∀ a ∈ Hask)

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,702 @@
...
+```haskell
+printState :: (Show s, Show a) => State s a -> s -> String
+printState (Pure x) s = "pure (" <> show x <> "," <> show s <> ")"
+printState (Free m) s =
+  let (x, s') = runStateF m s
+  in "state change " <> show s <> " -> "<> show s' <> "\n"
+    <> printState x s'
+```
+
+If we run `someComputateion` defined above through this interpreter, we'll get the following output:
```
alyoanton9: Typo: `someComputation`

Introduction to Free Monads/article.md
lierdakil: The point of this comparison is we can define a `Monad` instance with `fmap`, `return` and `join`. We already have `fmap` by the merit of requiring a `Functor` instance, and `Pure` and `Free` provide the encodings for `return` and `join` respectively. Perhaps I should've gone into a bit more detail here.

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,702 @@
...
+
+```
+state change 1 -> 1
```
alyoanton9, line 404: BTW, where is it defined that the initial state value is 1?

lierdakil: Oh, indeed, it isn't. "Free state" is exactly the same as regular "state", behaviour-wise, so the initial state is supplied to `runState` (or in this case `printState`). But I've neglected to mention the exact call I'm doing to produce the output, which is of course `printState someComputation 1`.

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,702 @@
...
+
+In general, we can restore any proper monad behavior from a free monad, since it doesn't actually define any semantics, beyond those necessary for any monad.
+`free` provides a function for that:
+
+```haskell
+foldFree :: Monad m => (forall x. f x -> m x) -> Free f a -> m a
+```
+
+Given a function[^natural-transformation] converting `f x` into some monad `m x` for any `x`, we can convert `Free f a` into `m a`.
+Returning to our list example, we can get the original list behavior by applying `foldFree id`.
```
alyoanton9: Perhaps it's obvious, but it could be convenient to have the original behaviour right here (just for presentation purposes)

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,702 @@
...
+          var <> " <- add " <> x <> " " <> y <> "\n"
+        go $ g var
+      Input x -> do
+        var <- lift freshVar
+        tell $ var <> " <- input\n"
+        go $ x var
+      Output s next -> do
+        tell $ "output " <> s <> "\n"
+        go next
+      If cond tru fls -> do
```
alyoanton9: **Nitpick**: why not just `true` and `false` or `onTrue`, `onFalse`?

lierdakil: Good point, `onTrue` and `onFalse` looks more self-explanatory.

If you care to know the reason why I went with `tru` and `fls`, that's something I've dragged up (not necessarily intentionally) from pure untyped lambda calculus, where the Church encoding of booleans commonly uses these identifiers.

alyoanton9:
### General impression
You might notice I didn't leave many comments on the examples part. I read through them several times and grasped the main idea of Free Monads and why we need them, but not the every single example. And I can't explain why, perhaps it's too complicated sometimes, perhaps I require more time to fully understand this.

For example, in this paragraph only the last sentence is 100% clear for me 😅 
> One caveat is that in the case of a non-branching computation, we only want our base functor to be a functor in the overall result type, but not in the arguments...

One more notice -- "too many" _caveat_ usages. Despite the fact there are 3 of them, I read it and felt like _oh, caveat again_. This is kind of nitpick, though. Probably, it sounds over-used for me only.

All in all, as you pointed in the conclusion, after this article the main idea and possible applications are pretty clear, so I believe all the examples above are not necessary fully understandable.
> after this brief introduction, you understand, at least in principle, what free monads are, and how they may be useful. Of course, the only way to get proficient with them is to use them in projects, and no blog post can replace actual hands-on experience

The writing style is also friendly for realizing 👍

The last, but not the least, I like the idea of using matching quotes, it makes the text less formal and, again, friendly and even funny in some cases.

Introduction to Free Monads/article.md
lierdakil: I'll explain my thought process here. So while writing this, I've realized I just can't find it in me to not quote the "mostly wrong history of programming languages", because I think it's brilliant, simultaneously funny and provides some deep insight into the underlying theory around monads. But then the only section with a quote stood out like a sore thumb, so I've decided to add relevant quotes to each section. The "free speech" vs "free beer" did come naturally right away. The other ones, not so much. So, for some definition of "relevant", this Linus quote is the most relevant to the "examples" section I could come up with (I also find it to be a little funny, so I feel it helps a bit with keeping the style more in the realm of "blog post" and less in the realm of "boring university lecture")

I do agree that the last two quotes don't bring much new to the table, but I also think dropping the quotes altogether halfway through would feel weird. I'm very open to suggestions for better quotes, however, so if you (or anyone else) has ideas, I'd very much like to hear those.

Introduction to Free Monads/article.md
lierdakil: I do go over the uses in the examples and in conclusion, so I don't know if it makes much sense repeating that here -- especially considering the reader likely has no idea of what free monads are yet.

Maybe I'm looking at this wrong, but I find it to be much more natural to first explain the concept in some detail, and then go over how it is useful. That way, the reader/listener can make their own judgement whether it is indeed useful, since they have at least some understanding at that point. Starting off with "here's how you can use that thing which you have no idea about" is rather prevalent, but I personally don't particularly like this approach. In my opinion, it's like starting the book from the back -- you read the payoff the author spent the whole book setting up, but you don't have enough context to appreciate (or sometimes even understand) it.

Introduction to Free Monads/article.md
lierdakil: :thinking: I could come up with some variations on the examples described here, but those would be somewhat boring. I could also trim the examples section somewhat, and present what I trim off as exercises, however that would diminish the value of the article on its own (i.e. for someone who doesn't care much for exercises). I'll see if I can come up with something more interesting, though.

FWIW, I do have one exercise in the middle of the theoretic section (i.e. equivalence proof of monad laws defined via bind vs Kleisli composition), but honestly I don't expect anyone to actually do it  :sweat_smile:

Introduction to Free Monads/article.md
lierdakil: > Comparing Cons with Free looks quite natural to me, as it encodes action followed by the sequence of actions

Also, about that... I briefly mention that `Free` is a bit more general than `Cons`, because, unlike `Cons`, it contains an arbitrary functor, which potentially makes it into a tree rather than a list.

So this intuition of "a sequence of actions" only holds as long as the underlying functor is non-branching, i.e. in terms of Haskell, each data constructor has at most one argument referencing the functor variable and the datatype isn't recursive. In the general case, `Free` encodes an arbitrary branching tree of actions, not a sequence (which should be evident in the later examples)

Perhaps I should go into a bit more detail in the text.

Introduction to Free Monads/article.md
NaeosPsy: I think that not every section needs to (necessarily) have a quote. So, if there is something that doesn't really add to the whole experience, it is ok to drop it & not strain oneself too much to find some quote, since (most of the) readers won't look at the article with the same kind of structural outlook. 

It is pretty much a judgement call though, and I won't be very unhappy if you decide to keep them.

Introduction to Free Monads/article.md
NaeosPsy: I don't think this explains free monads, right? 

```
If you're relatively new to Haskell, you might feel that monads are hard enough to grasp as is, and now there's these different type of monads to worry about.
Perhaps you've used Haskell for a while but didn't yet get to wrapping your head around free monads.
In any case, the concept is ultimately rather straightforward.
This article aims to introduce free monads and to explain why they are useful.
```

Basically, you want to make people be able to match with their goals as fast as possible. People frequently Google random stuff, and this will be the most frequent way that this article will be opened (along with other links from our blog). 

So your goal (and the chance this paragraph gives) is to say how free monads will matter to the person at the moment when they are Googling: you need to convince them to sit down, buckle up, and read the rest of the article. (And I don't mean convince in a persuasive sense, but in a more synergy-kind-of-way.) While a person that doesn't need to be convinced will probably just skip to the meat of the article, so none of your actions here will matter much.

If you say the above-mentioned four sentences, you refuse this chance, kinda.

Introduction to Free Monads/article.md
NaeosPsy: Boring is fine, usually exercises are at the end and they are very skippable, but whomever will need it -- implement a free monad instance for this or smth like that will be just in time. :)

lierdakil:
> For example, in this paragraph only the last sentence is 100% clear for me

FWIW, I rewrote the paragraph in question to hopefully make it more clear. That also had the side effect of removing the "caveat" from there.

Introduction to Free Monads/article.md
alyoanton9: Okay, I see, just mixed the sets (which one includes another) when I was reading the definition at Wikipedia. The example is clear too. However, I believe it would be nice to mention that embedding operation is injective (and perhaps include some more words to complete the definition). Despite the fact the "nature" of embedding becomes clear with example, the reader go through the text sequentially, so that it seems logical to proceed to examples when the definition is clear.

Introduction to Free Monads/article.md
alyoanton9: Okay, never mind, I just saw the analogues where they weren't meant.

I've read the paragraph one more time, it sounds pretty clear, though.

Introduction to Free Monads/article.md
alyoanton9: > from pure untyped lambda calculus, where the Church encoding of booleans commonly uses these identifiers.

Ah, I see 👌

Introduction to Free Monads/article.md
alyoanton9: Oops, you've already added the footnote, 🆒

Introduction to Free Monads/article.md
alyoanton9: > Basically, you want to make people be able to match with their goals as fast as possible. People frequently Google random stuff, and this will be the most frequent way that this article will be opened 

➕ 

I think it's a frequent scenario, when you want to implement X and remember (or heard, or assume) that it could be done with Y. Then you google Y and go through relevant links. If you see the article where the first sentences say something like "Y is commonly used for A, B, ... X"  and "we are going to show you what Y is", there is a high chance you will stay on the page. However, when there is only "Y is ..." and nothing about X, the chance is lower.

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,735 @@
...
+Let us start with a simple example, specifically, `Monoid`s.
+In Haskell, any type `α` in the `Monoid` type class must have two functions defined.
+First, to construct a neutral element, we have `mempty :: α`.
+Second, to in some sense combine two elements, we have `mappend :: α -> α -> α`.
+Additionally, these functions must satisfy three equations, called the monoid laws:
+
+```haskell
+(x `mappend` y) `mappend` z = x `mappend` (y `mappend` z) -- associativity
+mempty `mappend` x = x -- left identity
+x `mappend` mempty = x -- right identity
```
heitor-lassarote: I'd prefer if `<>` was used here instead, not only I think it's more readable, but from [`mappend`'s documentation](https://hackage.haskell.org/package/base-4.16.0.0/docs/Data-Monoid.html#v:mappend):

> In a future GHC release `mappend` will be removed from `Monoid`.

So using `<>` will be more future-proof.

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,735 @@
...
+```haskell
+type M = [S]
+```
+
+then, our embedding can be defined as
+
+```haskell
+i :: S -> M
+i x = [x]
+```
```
heitor-lassarote: As someone with not much background in pure mathematics, having examples in Haskell like this is a great aid to me. I'd assume most people reading this article may read the introduction and feel intimidated (I know I felt like so). After this example, I feel like I understand it now, but I still don't get what is meant by this part:

> $M$ only contains elements required to exist by $i$ or the operations on the elements of $M$.

Perhaps a paragraph elaborating a bit more on this would be nice.

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,735 @@
...
+(f >=> g) >=> h = f >=> (g >=> h) -- associativity
+return >=> f = f -- left identity
+f >=> return = f -- right identity
+```
+
+That is why a monad is indeed a monoid: it satisfies the same laws, and the "values" are of the type `Functor f => a -> f b`.
+Note, though, that this is not the category of endofunctors.
+However, but there exists a natural transformation to the category of endofunctors[^endofunctors].
+
+[^endofunctors]: One could alternatively define monads via `join :: m (m a) -> m a` instead of `>>=`, which would make it a little more evident that monads are monoids over endofunctors, but to explain this properly, we would need to use a bit more of category theory jargon.
```
heitor-lassarote: Once I saw in a Discord server a very good [discussion](https://discord.com/channels/280033776820813825/280036215477239809/928715018441203753) about this. First consider you have the following:

```hs
type Pair a b = (a, b)
type f ~> g = forall x. f x -> g x
newtype Identity x = Identity x
newtype Compose f g x = Compose (f (g x))
```

Then, you have:

```hs
class MonoidInHask m where
  mempty :: () -> m
  mappend :: Pair m m -> m
```
Then a monad may look like so:
```hs
class MonadInHask m where
  return :: Identity ~> m
  join :: Compose m m ~> m
```
And these two structures look alike. But here you talk about Kleisli arrows, which I think it's even easier to read.

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,735 @@
...
+If we instead rewrite these laws using the Kleisli composition
+
+```haskell
+(>=>) :: Monad m => (a -> m b) -> (b -> m c) -> (a -> m c)
+f >=> g = \x -> f x >>= g
+```
+
+it becomes much more apparent[^equivalence]:
+
+[^equivalence]: The proof of equivalence is left as an exercise to the reader.
```
heitor-lassarote: I think it would be great to add a point explaining what equivalence is.

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,735 @@
...
+```
+
+The first line of output corresponds to `get`.
+While it doesn't modify the state, it's still an action, so the `state change` line is still printed.
+The second line corresponds to `put $ i + 1`, which naturally changes the state to `2`.
+The third line corresponds to `pure ()`, but it also outputs the end state.
+
+Let's summarize what we've learned so far.
+We can take any usual base functor for some monad, and get a `Monad` instance "for free".
+The catch is that we need to define the semantics of this monad separately by writing an interpreter.
```
heitor-lassarote: I'm not sure about using "the catch" here, since it makes it seem like a negative connotation. The fact that you can interpret the same thing in different manners is one of the advantages of the free monads, as far as I understand.

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,735 @@
...
+The free monad doesn't specify the meaning of monadic actions, so we have to decide what those actions mean when we're running it.
+
+To illustrate this point, we're going to write another interpreter, which is essentially a pretty-printer for the flow of a `State` computation:
+
+```haskell
+printState :: (Show s, Show a) => State s a -> s -> String
+printState (Pure x) s = "pure (" <> show x <> "," <> show s <> ")"
+printState (Free m) s =
+  let (x, s') = runStateF m s
+  in "state change " <> show s <> " -> "<> show s' <> "\n"
```
heitor-lassarote:
```suggestion
  in "state change " <> show s <> " -> " <> show s' <> "\n"
```

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,735 @@
...
+For instance, you might know that the list monad encodes non-determinism.
+However, if we derive a free monad for lists, the behavior is slightly different.
+Consider:
+
+```haskell
+listComputation :: Free [] Int
+listComputation = do
+  x <- liftF [1, 2, 3]
+  y <- liftF [70, 80]
+  z <- liftF [500, 600]
```
heitor-lassarote: Nitpick: I think this example could be simplified a bit, perhaps using `[1, 2, 3]`, `[10, 20]` and `[100, 200]`, as those numbers are a bit easier to parse in my head. Then we get `[[[111,211],[121,212]],[[112,212],[122,222]],[[113,213],[123,223]]]`, which is easier to tell what's going on.

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,735 @@
...
+data ASTF t a
+  = Add t t (t -> a)
+  | Input (t -> a)
+  | Output t a
+  deriving Functor
+```
+
+Notice that for `Add` we're saying that it takes two arguments `t`, its result is also `t`, but ultimately, the continuation returns the type `a`.
+In practice, we will define `add :: t -> t -> Free (ASTF t) t`, but to have a lawful `Functor`, and also more flexibility, we're not assuming these types are the same.
+Similar reasoning applies to `Input`.
```
heitor-lassarote: I'd move this explanation up, where you talk about the "key intuition". At least, that wasn't very clear to me until I got here, and the reader may stare at `ASTF` thinking "but why is it like this?"

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,735 @@
...
+```haskell
+type FreeAST t = Free (ASTF t)
+
+input :: FreeAST t t
+input = liftF $ Input id
+
+add :: t -> t -> FreeAST t t
+add x y = liftF $ Add x y id
+
+output :: Show t => t -> FreeAST t ()
```
heitor-lassarote: Do we really need this constraint here?

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,735 @@
...
+The third line corresponds to `pure ()`, but it also outputs the end state.
+
+Let's summarize what we've learned so far.
+We can take any usual base functor for some monad, and get a `Monad` instance "for free".
+The catch is that we need to define the semantics of this monad separately by writing an interpreter.
+
+### List as a free monad
+
+Not all free monads behave exactly the same as their regular counterparts.
+For instance, you might know that the list monad encodes non-determinism.
```
heitor-lassarote, line 421: Does it? Could you explain this part, or link some resources for interested readers like me?

Introduction to Free Monads/article.md
heitor-lassarote: +1 on this, theory without exercises is not very useful, since the reader may rapidly forget about it, and not know how to come up with exercises by themselves.

Also, one thing that might be useful at the very bottom of the article are resources like "what next?", I was hoping you'd mention [freer monads](https://okmij.org/ftp/Haskell/extensible/more.pdf), or something about comonads.

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,735 @@
...
+```haskell
+liftF :: (Functor f, MonadFree f m) => f a -> m a
+```
+
+`MonadFree` is just an MTL-style type class for different encodings of the free monad.
+In our case, we can pretend that
+
+```haskell
+liftF :: Functor f => f a -> Free f a
+```
```
heitor-lassarote: Maybe you should define this, or say this is an exercise to the reader.

heitor-lassarote (commented):
Amazing article, the idea of free monads is explained very clearly. The concepts that I didn't know, I've learned here, and the concepts that I already knew, were refined. I even learned some new markdown tricks. :)

I've left some comments which I think will make the article even easier to follow. Thank you for writing this! o/

Introduction to Free Monads/article.md
lierdakil: It certainly does. Consider:

```haskell
listComputation :: [Int]
listComputation = do
  x <- [1, 2, 3]
  y <- [10, 20]
  pure $ x + y
-- the result is [11,21,12,22,13,23]
```

You can read this as "let x be a value from the set {1, 2, 3} and y be a value from the set {10, 20}. What are all possible results of the expression `x + y`?" Hence, nondeterminism. There are a few caveats to this (related to ordering and laziness) and continuation-based `ListT` or `LogicT` is more powerful, but thinking of the list monad as a non-deterministic computation works pretty well as a mental framework. You could also think of those as nested `foreach` loops of course.

Introduction to Free Monads/article.md
lierdakil: It's defined in `Control.Monad.Free` (albeit with a slightly more general signature). The definition is pretty trivial, and that's not really the point. So an exercise for the reader it is.

Introduction to Free Monads/article.md
lierdakil: I didn't really want to get into this whole thing with the `Semigroup`, but you're probably right, it's better to eschew `mappend`.

Introduction to Free Monads/article.md
lierdakil: > $M$ only contains elements required to exist by $i$ or the operations on the elements of $M$.

In other words, $M$ contains all elements, that could be produced by the embedding $i$ from elements of $S$, and any additional elements required for operations (defined as part of the free structure) on $M$ to be closed wrt $M$. The latter condition is required for operations Mⁿ → M to be total, for example, subtraction isn't total wrt natural numbers, `1 - 2` is undefined in ℕ. The concept is rather simple, but it's a bit hard to articulate concisely.

Introduction to Free Monads/article.md
lierdakil: Yes, but in context -- i.e. getting the Monad instance "for free" -- it basically is. As in, "I'll give you this beer for free, but you have to ferment it yourself" or something.

... It's an attempt at a joke.

Introduction to Free Monads/article.md
lierdakil: I actually can't move it up there, because I didn't at that point introduce the notion that we have to use _two_ type parameters. I can add a short footnote explaining what's going on though.

Introduction to Free Monads/article.md
lierdakil: Indeed, we don't. It's a holdover from an earlier version of the code (wherein I used `String` as a parameter of `Output`).

lierdakil:
@heitor-lassarote, I believe I've addressed your comments in the latest commits.

I also did add a paragraph on freer monads to the conclusion, among other things.

@NaeosPsy, I've tried to rewrite the introduction, and I've added a few exercises. I've also removed the two quotes I was on the fence about. I could also drop the Dennis Ritchie one, but I feel like it adds a little to the overall reading experience, not so much the quote itself, but the mere fact that it's there.

I've also found a couple typos in the examples, so those are fixed.

Finally, I've added a short-ish subsection on decomposition approaches wrt ASTs. It's all pretty obvious stuff, but it occurred to me it might be good to at least mention that all of this is easily decomposable.

Side note, I could post all the examples here to github somewhere, but not entirely sure where. Is there a best practice for this kind of thing, or do I publish it wherever and just link that?

NaeosPsy:
@lierdakil I do not entirely understand what you mean by publishing examples. Is it regarding exercises?

We usually use `<details>` HTML tags (with a `<hr>` at the end of the inside text for readability) for giving exercise solutions. 

Full code of code example can be uploaded as a GH Gist and linked or put in a `<details>` tag as well.

lierdakil:
Uh. I mean publishing all the code I have in this here article in compilable/runnable form (well, "runnable" with `stack repl`).

Plopping full code under \<detalis> wouldn't be _too_ convenient, since there are some naming conflicts.

lierdakil:
Yeah, gist (or more likely a repo) on my personal account would fall under "publish wherever and link it".

Introduction to Free Monads/article.md
heitor-lassarote: I think I understand now, thank you for taking your time explaining this to me.

heitor-lassarote (approved):
I haven't resolved the exercises myself, but they look pretty good. I think this post is great. Thank you for taking a look at my comments!

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,892 @@
...
+
+Now, `List a` is a type, but `Free f` is a unary type constructor, so we'd need one more type argument.
+Doing that directly yields
+
+```haskell
+data Free f a = Nil a | Cons f (Free f) a
+```
+
+We can't define such a data type, due to `f` not being a plain type and hence can't be an argument of a data constructor, but we're actually very close.
```
NaeosPsy:
```suggestion
We can't define such a data type due to `f` not being a plain type (hence, it can't be an argument of a data constructor), but we're actually very close.
```

Or you can rephrase in another way, but: no comma after due and there is something odd in the original sentence around hence.

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,892 @@
...
+
+Let us now look at the actual definition of `Free f`:
+
+```haskell
+data Free f a = Pure a | Free (f (Free f a))
+```
+
+As you can see, it is indeed very similar to a list.
+Note, however, that it is a bit more general: the functor is arbitrary, potentially making it a tree, with branches of type `f` and leaves of type `a`.
+Pure values, i.e. leaves, are encoded via the constructor `Pure`, and "actions", i.e. branches -- via the constructor `Free`.
```
NaeosPsy:
```suggestion
Pure values (leaves) are encoded via the `Pure` constructor, and "actions" (branches) via the `Free` constructor.
```

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,892 @@
...
+```haskell
+printState :: (Show s, Show a) => State s a -> s -> String
+printState (Pure x) s = "pure (" <> show x <> "," <> show s <> ")"
+printState (Free m) s =
+  let (x, s') = runStateF m s
+  in "state change " <> show s <> " -> " <> show s' <> "\n"
+    <> printState x s'
+```
+
+If we run `someComputation` defined above through this interpreter, for example by invoking `printState someComputation 1`, we'll get the following output:
```
NaeosPsy: For example needs to be surrounded by commas. But I think it's easier to just take it out, don' t think it does anything here.

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,892 @@
...
+Let's summarize what we've learned so far.
+We can take any usual base functor for some monad, and get a `Monad` instance "for free".
+There's no free lunch here though, so we need to define the semantics of this monad separately by writing an interpreter (or several if we need to).
+
+### List as a free monad
+
+Not all free monads behave exactly the same as their regular counterparts.
+For instance, you might know that the list monad encodes non-determinism[^non-determinism].
+However, if we derive a free monad for lists, the behavior is slightly different.
+Consider:
```
NaeosPsy: Just "consider" doesn't sound very nice.

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,892 @@
...
+Of course, we can get the standard list monad behavior by concatenating the nested lists, but we might not necessarily want to do that.
+
+In general, we can restore any proper monad behavior from a free monad, since it doesn't define any semantics beyond those necessary for any monad.
+`free` provides a function for that:
+
+```haskell
+foldFree :: Monad m => (forall x. f x -> m x) -> Free f a -> m a
+```
+
+Given a function[^natural-transformation] converting `f x` into some monad `m x` for any `x`, we can convert `Free f a` into `m a`.
```
NaeosPsy: I think there should be a comma before "for any x"?

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,892 @@
...
+      Input x -> do
+        var <- lift freshVar
+        tell $ var <> " <- input\n"
+        pure $ x var
+      Output s next -> do
+        tell $ "output " <> s <> "\n"
+        pure next
+```
+
+Notice, up to this point, all constructors of our `ASTF` had exactly one parameter that had anything to do with the functor.
```
NaeosPsy, line 559:
```suggestion
Notice, up to this point, all `ASTF` constructors had exactly one parameter that had anything to do with the functor.
```

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,892 @@
...
+```haskell
+data ASTF t a
+  = Add t t (t -> a)
+  | Input (t -> a)
+  | Output t a
+  | If t a a
+  deriving Functor
+```
+
+We add a new constructor, `If`, with one scalar parameter and two branches.
```
NaeosPsy, line 573:
```suggestion
We add a new `If` constructor with one scalar parameter and two branches.
```

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,892 @@
...
+
+In abstract algebra, we call something a "free X" when it is in some sense a minimal structure that satisfies conditions for being X.
+The exact definition of "minimal" in abstract algebra is a little vague, but the main idea is that "free X" satisfies all laws for X, and does not have any other laws describing it.
+That is, in this context, "free" means "unrestricted" -- it is not that we can get a structure "for free"[^for-free], but that no other laws are restricting the structure apart from those absolutely necessary.
+
+[^for-free]: Although in most cases, we actually might, depending on your definition of "for free".
+
+More formally, a free structure over a set $S$ is a set $M,$ together with operations on elements of $M,$ such that:
+
+- There is an embedding[^embedding] $i: S\to M$.
```
NaeosPsy: Unfortunately, our blog engine doesn't support Markdown footnotes.

NaeosPsy:
@lierdakil I've submitted a commit with grammar edits (such as comma after i.e., taking out commas between and where it is not necessary, etc.) 

I've also commented in some places where I think it is important you look at the suggestion before it is commited. 

The biggest problem is that the engine doesn't render Markdown footnotes. Since I don't think we can add them that fast, it would be good if you could adapt your article for that.

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,894 @@
...
+```
+
+That is why a monad is indeed a monoid: it satisfies the same laws, and the "values" are of the type `Functor f => a -> f b`.
+Note, though, that this is not the category of endofunctors.
+However, there exists a natural transformation to the category of endofunctors[^endofunctors].
+
+[^endofunctors]: One could alternatively define monads via `join :: m (m a) -> m a` instead of `>>=`, which would make it a little more evident that monads are monoids over endofunctors, but to explain this properly, we would need to use a bit more of category theory jargon.
+    There is an excellent [r/math post][endofunctors] on the subject.
+
+    [endofunctors]: https://www.reddit.com/r/math/comments/ap25mr/a_monad_is_a_monoid_in_the_category_of/
```
NaeosPsy: For some reason gets rendered like this 
image

```suggestion
[endofunctors]: (https://www.reddit.com/r/math/comments/ap25mr/a_monad_is_a_monoid_in_the_category_of/)
```

Introduction to Free Monads/article.md
```diff
@@ -0,0 +1,894 @@
...
+
+Not all free monads behave exactly the same as their regular counterparts.
+For instance, you might know that the list monad encodes non-determinism[^non-determinism].
+However, if we derive a free monad for lists, the behavior is slightly different.
+Consider:
+
+[^non-determinism]: If the statement about list monad encoding non-determinism is surprising, consider that one could use a list to represent all possible outcomes of some event and then handle those one by one.
+    For further reading on this topic, refer to the [School of Haskell article on the list monad][school-of-haskell] or the [Haskell wikibook section on understanding monads][haskell-wikibook]
+
+    [school-of-haskell]: https://www.schoolofhaskell.com/school/starting-with-haskell/basics-of-haskell/13-the-list-monad#non-deterministic-computations
```
NaeosPsy: Same thing as in  https://github.com/serokell/blog-posts/pull/8/files#r789621026, or just use regular hyperlinks.

Introduction to Free Monads/article.md
NaeosPsy: Or just use a regular hyperlink.

lierdakil: On that note, what engine are we using? I've assumed pandoc-compatible Markdown, but apparently not. Is there a straightforward way to get a preview locally?

NaeosPsy: I don't know which engine exactly we use. I think @kirelagin would be able to answer this? 

There's probably not _a straightforward_ way to preview locally at this point of time. 🤔 All I can suggest is to use the admin panel of website's staging to submit the post and see what happens. 😅

alyoanton9 (approved):

Introduction to Free Monads/article.md
lierdakil: :thinking: I don't mind rephrasing here, but parentheses feel a bit odd. I didn't at this point clarify that branches and leaves correspond to "actions" and pure values, respectively. Hence, `i.e.` is used for emphasis. With parentheses, this feels more of an off-hand remark, which it in fact isn't.

Perhaps I should just split this into two sentences.

Introduction to Free Monads/article.md
lierdakil: Uh... Me not being a native speaker can make cases like this a bit hard.

So, what I'm _trying_ to say here is this: given a function `f : ∀ x. f x → m x`, we can get `g : ∀ a. Free f a -> m a`. Hence, "for any `x`" isn't a qualifier, it's what's defining the "given function". With a comma, it sounds to me more like "for any `x`" qualifies the second sentence, i.e. "we can convert", which isn't at all the intended meaning.

Introduction to Free Monads/article.md
NaeosPsy: I put parenthesis because you technically need comma after i.e. as well. Having it be `,i.e., leaves,` seemed quite horrible to read, while this looked quite nice. 

Having it be two sentences seems quite okay as well.

Introduction to Free Monads/article.md
NaeosPsy: Mhm. Yeah, then it is not necessary.

lierdakil:
@NaeosPsy, okay, I think I've managed to get rid of the footnotes (mostly replacing those with `<details>`). I'd like to think I've also addressed your grammar and phrasing concerns, but probably not really :sweat_smile: 

I've added GitHub links with examples and solutions to exercises (those are actually the same link, solutions are just in a different branch). I'm a bit on the fence if it's a good idea to provide the latter or not, though, but maybe that's my teaching experience speaking (i.e. I know for a fact most of my students won't even read the description if a solution is readily available :sweat_smile:)

Finally, I've realized I forgot to mention one technical bit related to different encodings of free monads -- that's two sentences in e2f2bd87424bc2bbc5366f324839efc5035d49d2.

I _think_ the formatting should be fine, I checked on the site via the in-editor preview, but it's not the same as the rendered article on the site itself afaict.

lierdakil:
Oh, right, lest I forget. Any suggestions on the licence I should put on https://github.com/lierdakil/free-monad-examples? Since I wrote that while on the clock, technically it's © Serokell, so I thought I should probably ask. In this particular case, I'd be inclined to use something like <https://unlicense.org/>, or, failing that, MIT.

NaeosPsy:
This question isn't really sorted. 🤔

There's this thing where we would want attribution if somebody uses content from Serokell Blog, but software licenses don't really talk about that, AFAIK. I think for this particular example, it is okay if you use a permissible license.

Introduction to Free Monads/article.md
kirelagin: I’m let to the party but we use some generic javascript markdown renderer (not that there is some standard anyway).

kirelagin: The problem in the highlighted code fragment is that there is something weird with the indentation? I don’t think reference links like `[...]:` can be indented like that.

kirelagin: In case you are really wondering, it’s this one: https://github.com/markdown-it/markdown-it

lierdakil: > I don’t think reference links like `[...]:` can be indented like that.

Oh, they absolutely can, at least in some flavours -- because it's in the footnote -- anything indented 4 spaces after the footnote is part of the footnote, and at least Pandoc's footnotes allow arbitrary block content. FWIW this code snippet works as expected with markdown-it-footnote, too.

PR #11, opened 2022-02-07
[PR-1246] Make corrections to Introduction to Free Monads

### LINKED ISSUE #10: Error in the "Introduction to Free Monads" blog post
@lierdakil This was a [Reddit comment](https://www.reddit.com/r/haskell/comments/shulfo/comment/hvkikhg/?utm_source=share&utm_medium=web2x&context=3), but then I found this repo and decided it's better to create an issue.

The following sentence in the blog post [Introduction to Free Monads](https://github.com/serokell/blog-posts/blob/master/Introduction%20to%20Free%20Monads/article.md) is wrong.

> On the other hand, addition over non-negative integers doesn’t constitute a free monoid. That would imply the law of commutativity, which isn’t a monoid law.

Non-negative integers under addition are a free monoid (which is isomorphic to the list monoid described in the blog post, given that the generator set `S` is a singleton). Because there is a single generator (so any element is of the form `s^n`), the commutativity relation follows from associativity.

The intuition _"The only laws that hold for the generated structure are those required to hold."_ is right, but as long as you don't want to get into the details of generators and relations, it's perhaps better expressed in terms of homomorphisms (i.e. structure-preserving functions: `f(x<>y) = f(x)<>f(y)`). If `M` is a free monoid over `S`, then for any monoid `X` and any function `f: S -> X` we can extend this function to a homomorphism `f': M -> X`. (And the minimality condition basically says that there is only one such homomorphism.)

---------------

I'm also not sure if "vague" is the best word here:

> The exact definition of “minimal” in abstract algebra is a little vague

I believe the exact definition in abstract algebra is rather precise. It's the article that does some handwaving (which I don't see as anything wrong by itself).

lierdakil:
Problem: Non-negative integers under addition are a free monoid,
but the article claims otherwise.

Solution: Remove "non-negative", expand a little on what is meant
by this, and add a details section explaining why this was an
error and how removing "non-negative" fixes it.

Resolves #10 (hopefully)

@kukimik, I'd appreciate it if you took a quick look to make sure I didn't accidentally write something wrong again :sweat_smile:

### DIFF
--- a/Introduction to Free Monads/article.md
+++ b/Introduction to Free Monads/article.md
@@ -64,7 +64,7 @@ x <> mempty = x -- right identity
 ```
 
 A free monoid thus has to have these two operations (together with an embedding of some underlying set) and satisfy these laws and only them.
-For example, it is implied that commutativity, i.e. the equation `x <> y = y <> x`, should only hold if `x = y` or `x = mempty`, or `y = mempty`, and not in general.
+For example, it is implied that commutativity, i.e. the equation `x <> y = y <> x`, should only hold if it follows from the monoid laws.
 
 To give a more concrete example, a list (or, more generally, a sequence) of elements of the set `S` constitutes a free monoid over `S`.
 Indeed, if the set $M$ is a set of lists of elements of $S$:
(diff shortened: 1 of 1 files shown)


Introduction to Free Monads/article.md
```diff
@@ -84,8 +84,42 @@ Then `mempty = []` and `(<>) = (++)`.
...
-That would imply the law of commutativity, which isn't a monoid law.
+On the other hand, addition over integers doesn't constitute a free monoid.
+That would imply the law of commutativity, which isn't a monoid law, and, in case of integers, doesn't follow from monoid laws.
+
+<p>
+	<details>
+	<summary>On non-free monoids</summary>
+
+The previous version of this article incorrectly claimed that addition over _non-negative_ integers doesn't constitute a free monoid (thanks Michał Kukieła for the correction!).
+The reason for why that is incorrect is that for naturals, commutativity immediately follows from associativity:
```
kukimik, line 95: Note that this contradicts line 67 (and that line 67 is therefore wrong):

> For example, it is implied that commutativity, i.e. the equation `x <> y = y <> x`, should only hold if `x = y` or `x = mempty`, or `y = mempty`, and not in general.

Introduction to Free Monads/article.md
```diff
@@ -84,8 +84,42 @@ Then `mempty = []` and `(<>) = (++)`.
...
+\\ & = k + n
+\end{align}$
+
+This doesn't hold if integers can be of different sign.
+To see why, assume, e.g. $n < 0 < k:$ in this case, it's impossible to get $k+n$ from $n+k$ just by reshuffling parentheses.
+
+Another way to look at it is this: a free monoid doesn't do anything interesting.
+Hence, we can "recover" any other monoid from a free one.
+
+More formally, if $M$ is a free monoid over $S$, then for any monoid $N$, given a map $f : S \to N$, we can extend this mapping to a monoid morphism $f' : M \to N.$
```
kukimik:
```suggestion
More formally, if $M$ is a free monoid over $S$, then for any monoid $N$, given a map $f : S \to N$, we can extend this mapping to a monoid morphism $f' : M \to N$ in a unique way.
```

The uniqueness is needed for the implication in line 112. I think it does not hurt to add it.

Introduction to Free Monads/article.md
```diff
@@ -84,8 +84,42 @@ Then `mempty = []` and `(<>) = (++)`.
...
+
+This also implies that all free monoids over the same set are isomorphic.
+
+Thus, if addition with naturals is a free monoid over some set $S$, then we can convert this to a list monoid over the same set (which we know is a free monoid).
+The trick is in the choice of a set: if we choose $S = \lbrace 1 \rbrace,$ we can map addition to concatenation, and natural numbers to lists of corresponding lengths.
+This mapping is unique, because $S$ is a singleton set.
```
kukimik: I'd remove this line. I think the argument is clear without it.

kukimik (commented):
Ha, that's my first review on github. :smile: 

In general it looks ok to me. I just made some minor comments.

The integers example is a bit tricky and may be a little hard to grasp  (perhaps because the integers are free when viewed as a group/abelian group). Personally I'd probably go with a simpler example: `{True,False}` with `AND` (or perhaps integers modulo n with addition). There you have the relations `x^2=x` (or `x^n = mempty`) which clearly can't follow from associativity. On the other hand, for the same reason the integers are also more interesting, so by no means am I suggesting to change this. This is just a general afterthought.

lierdakil:
@kukimik, thank you very much for the review. I've amended line 67 a bit, and implemented your other suggestions in [3bf978a](https://github.com/serokell/blog-posts/pull/11/commits/3bf978a0b519d52f4fca2c8d49a0591a71be27b5).

PR #12, opened 2022-04-11
[PR-1320] Article about untyped lambda calculus

lierdakil:
This should (hopefully) be ready for review, but I might give it another read and pick some nits later this week.

### DIFF
--- a/untyped-lambda-calculus/article.md
+++ b/untyped-lambda-calculus/article.md
@@ -47,27 +47,28 @@ The initial formulation had a logical inconsistency known as [Kleene–Rosser pa
 To sidestep this issue, Church isolated the part of lambda calculus relevant only to computation in 1936.
 This isolate is now known as the untyped lambda calculus.
 
-Later, In 1940, a typed version of lambda calculus based on Russel's type theory was introduced. 
+Later, in 1940, a typed version of lambda calculus based on Russel's type theory was introduced.
 It has weaker expressive power but it's logically consistent.
 
 [Kleene–Rosser paradox]: https://en.wikipedia.org/wiki/Kleene%E2%80%93Rosser_paradox
 [paradoxes-contemporary-logic]: https://plato.stanford.edu/entries/paradoxes-contemporary-logic/#IncoCertFormLogi
-
-The lambda symbol used in the standard notation, from which lambda calculus derives its name, most likely was chosen arbitrarily.
-However, there's an apocryphal story the notation was derived from $\hat x$ morphing first into $\Lambda x$ and later into $\lambda x$.
 
 In the mid-1960s, Peter Landin showed that lambda calculus models arbitrarily complex programming languages.
 This insight arguably kickstarted the research on functional programming languages.
 
 ## The pure untyped lambda calculus
 
-The untyped lambda calculus is, first and foremost, a consistent expression rewriting system.
-Calculus that only contains functions and variables is called the pure untyped lambda calculus, "pure" meaning it doesn't have anything beyond what's strictly necessary.
+The simplest, smallest type of lambda calculus is the pure untyped lambda calculus, so we'll start with it.
+
+The untyped lambda calculus is, first and foremost, a consistent system for rewriting expressions.
+Lambda calculus that only contains functions and variables is called pure, meaning it doesn't have anything beyond what's strictly necessary.
 
 We can formally define the pure untyped lambda calculus in terms of _abstractions_ and _applications_.
 
-_Abstraction_, or functional abstraction, is a definition of a parametric expression, that is to say, a function.
-Lambda calculus only defines univariate functions, but that's easily extended to multivariate, as we'll see later.
+### Abstraction
+
+An _abstraction_, or a functional abstraction, is a parametric expression, that is to say, a function.
+Lambda calculus only defines univariate functions, but we can easily extend it to include multivariate functions too, as we'll see later.
 
 We introduce new abstractions using the $\lambda$ symbol.
 The abstraction consists of the _head_ and the _body_, separated by the dot ($.$) symbol.
@@ -78,6 +79,21 @@ For example, $\lambda x. x$ is an abstraction.
 As usual in abstract mathematics, we can assign a name to an expression.
 Here we'll be using the equals sign ($=$) for that.
 For example: $id = \lambda x. x.$
+
+Lambda calculus derives its name from the $\lambda$ symbol used in this notation.
+The symbol was most likely chosen arbitrarily.
+However, there's an apocryphal story the notation was derived from $\hat x$ morphing first into $\Lambda x$ and later into $\lambda x$.
+
+Variables in the body of an abstraction that are parameters of said abstraction are called _bound variables_.
+Other variables are called _free variables_.
+
+An expression without free variables is called a _closed term_ or a _combinator_.
+
+The choice of bound variable names is arbitrary.
+Expressions that differ only in the names of bound variables are called _alpha-equivalent_.
+For example, $\lambda x. x$ and $\lambda y. y$ are alpha-equivalent, but $\lambda x. y$ and $\lambda y. z$ are not.
+
+### Application
 
 _Application_ is the operation of substituting a specific value for the parameter in the abstraction.
 It is the only operation defined in the pure untyped lambda calculus.
@@ -93,18 +109,9 @@ Since application binds more tightly, we don't need parentheses around the body 
 However, we might need parenthesis around the whole abstraction itself.
 We will also eschew parentheses for nested abstractions, e.g. $\lambda x. \lambda y. x\;y$ is $\lambda x. (\lambda y. (x\;y)).$
 
-Variables in the body of an abstraction that are parameters of said abstraction are called _bound variables_.
-Other variables are called _free variables_.
+## Defining computation
 
-An expression without free variables is called a _closed term_ or a _combinator_.
-
-The choice of bound variable names is arbitrary.
-Expressions that differ only in the names of bound variables are called _alpha-equivalent_.
-For example, $\lambda x. x$ and $\lambda y. y$ are alpha-equivalent, but $\lambda x. y$ and $\lambda y. z$ are not.
-
-## Defining the computation
-
-The step of the computation is called beta-reduction.
+A step of computation in lambda calculus is called beta reduction.
 It is a single application of one expression to another.
 Formally, beta-reduction uses the following rules.
 Given an expression of the form $(\lambda x. t_1)\; t_2$ (i.e. an application of some abstraction to some expression $t_2$):
@@ -127,6 +134,8 @@ Then we can do the rest of the steps with no fear of confusion: $\lambda z. (\la
 
 In the last example, you might wonder if we should also apply beta-reduction inside the abstraction body.
 It brings us to the discussion of evaluation strategies.
+
+### Evaluation strategies
 
 Evaluation strategy is a rule that defines which redexes are reduced and in what order.
 
@@ -153,33 +162,6 @@ It's interesting to note that the call-by-value strategy corresponds to eager ev
 In eager evaluation, arguments are evaluated before the function call.
 Call-by-name, on the other hand, corresponds to lazy evaluation.
 In lazy evaluation, some arguments are potentially never evaluated at all.
-
-To get a feel for how the weak strategies work, let's consider a small example:
-
-$(\lambda x. \lambda y. \lambda z. x\; x\; y)\; ((\lambda z. z)\; a)\; ((\lambda w. w)\; b)\; ((\lambda p. p)\; c)$
-
-Under call-by-value strategy, we evaluate arguments before application:
-
-$(\lambda x. \lambda y. \lambda z. x\; x\; y)\; ((\lambda z. z)\; a)\; ((\lambda w. w)\; b)\; ((\lambda p. p)\; c)$ \
-$\to (\lambda x. \lambda y. \lambda z. x\; x\; y)\; a\; ((\lambda w. w)\; b)\; ((\lambda p. p)\; c)$ \
-$\to (\lambda y. \lambda z. a\; a\; y)\; ((\lambda w. w)\; b)\; ((\lambda p. p)\; c)$ \
-$\to (\lambda y. \lambda z. a\; a\; y)\; b\; ((\lambda p. p)\; c)$ \
-$\to (\lambda z. a\; a\; b)\; ((\lambda p. p)\; c)$ \
-$\to (\lambda z. a\; a\; b)\; c$ \
-$\to a\; a\; b$
-
-Under call-by-name strategy, we first evaluate applications, instead:
-
-$(\lambda x. \lambda y. \lambda z. x\; x\; y)\; ((\lambda z. z)\; a)\; ((\lambda w. w)\; b) ((\lambda p. p)\; c)$ \
-$\to (\lambda y. \lambda z. ((\lambda z. z)\; a)\; ((\lambda z. z)\; a)\; y)\; ((\lambda w. w)\; b)\; ((\lambda p. p)\; c)$ \
-$\to (\lambda z. ((\lambda z. z)\; a)\; ((\lambda z. z)\; a)\; ((\lambda w. w)\; b))\; ((\lambda p. p)\; c)$ \
-$\to ((\lambda z. z)\; a)\; ((\lambda z. z)\; a)\; ((\lambda w. w)\; b)$ \
-$\to a\; ((\lambda z. z)\; a)\; ((\lambda w. w)\; b)$ \
-$\to a\; a\; ((\lambda w. w)\; b)$ \
-$\to a\; a\; b$
-
-Note how in call-by-name we didn't have to compute $((\lambda p. p)\; c)$ argument, because it disappeared when applying.
-On the other hand, we had to compute $((\lambda z. z)\; a)$ twice.
 
 A practically significant variation on call-by-name is _call-by-need_, which is, at least in theory, used in Haskell.
 Call-by-need behaves exactly like call-by-name, except that reduction steps that would duplicate computations, don't.
@@ -238,13 +220,14 @@ Multivariate abstractions, naturally, can be partially applied, e.g. $(\lambda x
 ### Church booleans
 
 Let us now introduce booleans.
-We will follow Church here.
-However, it's not the only way.
+We will do it in the same way Church did in his original paper.
+However, that's not the only way.
 Arguably, there are infinitely many ways to encode values as functions.
 
 So, let us declare two combinators:
 $tru = \lambda x\; y. x,$ and $fls = \lambda x\;y. y.$
-We're calling these $tru$ and $fls$ to disambiguate them from atomic boolean values.
+We're calling these $tru$ and $fls$ to disambiguate them from actual boolean values $true$ and $false$ that aren't actually abstractions.
+$true$ and $false$ are, hopefully obviously, not a part of the pure lambda calculus, but we can introduce them as terms in an impure lambda calculus.
 
 Now, the structure of these definitions is rather curious.
 $tru$ is a two-argument function that returns its first argument and ignores the second, while $fls$ returns the second and ignores the first.
@@ -296,7 +279,7 @@ Defining other boolean combinators is reasonably straightforward by analogy.
 
 ### Church pairs
 
-Since we have booleans now, we can also encode pairs (and by extension, tuples as trees or right combs) as functions, returning one value when given $tru$ as an argument, and another when given $fls$.
+Since we have booleans now, we can also encode pairs (and by extension, tuples) as functions, returning one value when given $tru$ as an argument, and another when given $fls$.
 
 First of all, let us define a pair constructor combinator:
 
@@ -321,13 +304,13 @@ $\to tru\;x\;y$
 $= (\lambda a\; b. a)\;x\;y$
 $\Rightarrow x.$
 
-Here, double arrow $\Rightarrow$ signifies more than one step of beta-reduction.
+Here, double arrow $\Rightarrow$ signifies multiple steps of beta-reduction.
 
 ### Church naturals
 
 Since we encoded both pairs and booleans, we already can encode naturals simply as sequences of booleans, that is, as binary numbers.
 In practice, this turns out to be cumbersome, however.
-Instead, we'll define naturals similar to the Peano construction.
+Instead, we'll define naturals similar to the [Peano construction](https://wiki.haskell.org/Peano_numbers).
 Essentially, we'll be choosing an encoding for zero and then encoding other numbers as some successor function applied to zero.
 
 Church used the following definitions:
@@ -353,28 +336,30 @@ $succ = \lambda n. \lambda s\; z. s\;(n\;s\;z).$
 In the same manner, we can define addition and multiplication:
 
 $plus = \lambda m\; n. \lambda s\; z. m\;s\;(n\;s\;z),$
+
 $mul = \lambda m\; n. \lambda s\; z. m\;(n\;s)\;z.$
 
 The former counts $m$ starting from $n$, and the latter counts $m$ times $n$ starting from $z.$
 
 We can also easily test whether a value corresponds to zero (i.e. $c_0$) or not:
-$$isZero = \lambda n. n\;(\lambda x. fls)\;tru.$$
+
+$isZero = \lambda n. n\;(\lambda x. fls)\;tru.$
 
 We run into a bit of an issue when trying to define subtraction or a predecessor function.
 Turns out it's doable but neither pretty nor efficient.
 We'll use pairs of values containing two subsequent naturals.
-That way, we can stagger the counter by one.
+That way, we can have a counter that's lagging by one.
 To simplify the notation, we'll introduce two named helper abstractions:
 
 $zz = pair\;c_0\;c_0$\
 $ss = \lambda nn. pair\;(snd\; nn)\;(succ\;(snd\;nn)).$\
 
 The first element of the pair here is the predecessor, and the second is the successor.
-$zz$ encodes the starting point, and $ss$ is the staggered successor.
+$zz$ encodes the starting point, and $ss$ is the lagging successor.
 
 Now we can define the predecessor function:
 
-$$pred = \lambda n. fst\; (n\; ss\; zz).$$
+$pred = \lambda n. fst\; (n\; ss\; zz).$
 
 With this definition, $pred\;c_0 = c_0,$ which isn't ideal.
 However, we don't have an encoding for negative numbers, so we don't have much of a choice.
@@ -388,7 +373,7 @@ Such expressions don't have a normal form, and we call those _diverging_.
 
 The simplest diverging expression is the application of the so-called omega combinator to itself:
 
-$$\Omega = \lambda x. x\;x.$$
+$\Omega = \lambda x. x\;x.$
 
 Using any typical evaluation strategy, one step of beta-reduction of $\Omega\;\Omega$ gives us $\Omega\;\Omega$ again.
 
@@ -396,16 +381,16 @@ The omega combinator is somewhat useless on its own.
 However, we can extend it to be a little more useful.
 In particular, we can encode recursion using the fixed point combinator, also known as $Y$ combinator:
 
-$$Y = \lambda f. (\lambda x. f\;(x\;x))\;(\lambda x. f\;(x\;x)).$$
+$Y = \lambda f. (\lambda x. f\;(x\;x))\;(\lambda x. f\;(x\;x)).$
 
 A shorter equivalent form is sometimes cited:
 
-$$X = \lambda f. \Omega\; (\lambda x. f\;(x\;x)).$$
+$X = \lambda f. \Omega\; (\lambda x. f\;(x\;x)).$
 
 We should note that it's not the only fixed-point combinator.
 For instance, another famous one is the Turing's combinator (named after its discoverer):
 
-$$\Theta = (\lambda x\;y.y\; (x\; x\; y))\; (\lambda x\;y.y\; (x\; x\; y)).$$
+$\Theta = (\lambda x\;y.y\; (x\; x\; y))\; (\lambda x\;y.y\; (x\; x\; y)).$
 
 The idea here is to give us the ability to express recursion by passing the function as its first argument.
 The omega combinator infinitely replicates itself.
@@ -438,7 +423,8 @@ However, if we tried to apply call-by-value, we would get stuck pretty quickly t
 There are fixed-point combinators that work with call-by-value as well.
 One particular example is:
 
-$$Y_v = \lambda f. (\lambda x. f\;(\lambda y. x\;x\;y)) \;(\lambda x. f\;(\lambda y. x\;x\;y)),$$
+$Y_v = \lambda f. (\lambda x. f\;(\lambda y. x\;x\;y)) \;(\lambda x. f\;(\lambda y. x\;x\;y)),$
+
 which is basically the $Y$ combinator with additional abstractions inserted in the middle.
 
 ## Conclusions
@@ -484,22 +470,35 @@ One has to appreciate the elegance at least.
 
 6. Implement a pure untyped lambda calculus interpreter in your favorite programming language.
 
-    One question you might have is how to handle name collisions and encode alpha-equivalence.
-    There can be many approaches to this problem, but one of the more common ones is using the de Bruijn encoding.
-    The idea is to represent variables without naming them.
-    Instead, we can encode variables by their de Bruijn indexes, i.e. a reference to the binding abstraction represented as a natural.
-    The natural number is the "nesting level" of the variable relative to its binding abstraction, starting at $0$.
-    For example, $\lambda x. x$ can be encoded as $\lambda. 0,$ and $\lambda x\; y. x\;x\;y$ as $\lambda.\lambda. 1\;1\;0.$
-    Two terms are alpha-equivalent if and only if their de Bruijn representation is the same.
-    We should note that, during beta-reduction, de Bruijn indexes of free variables will change, but they will change predictably.
-    Consider, for example, the following term:
-    $$\lambda. (\lambda. \lambda.\lambda. 1\;(2\;1\;0))\; 0$$
-    (this is $\lambda x.succ\;x$).
-    It is quite apparent that $\lambda. \lambda.\lambda. 1\;(0\;1\;0)$ is not equivalent (injecting variable names back gives $\lambda x. \lambda s.\lambda z. s\;(z\;s\;z),$ when it should be $s\;(x\;s\;z)$).
-    But, you might notice that $\lambda. \lambda.\lambda. 1\;(2\;1\;0)$ is (by the rule of eta-reduction).
-    This line of reasoning gives us the intuition for the general rule.
-    When an expression is substituted, de Bruijn indexes of its free variables must increase by the de Bruijn index of the variable the expression replaces.
-    It should also be noted that when the left term contains free variables, those must be reduced by one, because application "strips" one level of $\lambda$, e.g.
-    $$\lambda. (\lambda. \lambda.\lambda. 1\;(2\;3\;0))\; 0$$
-    is $\lambda. \lambda.\lambda. 1\;(2\;2\;0),$
-    not $\lambda. \lambda.\lambda. 1\;(2\;3\;0)$ (indeed in this example $3$ references a nonexistent abstraction).
+<p>
+<details>
+<summary>Tips on handling name collisions</summary>
+
+One question you might have is how to handle name collisions and encode alpha-equivalence.
+There can be many approaches to this problem, but one of the more common ones is using the de Bruijn encoding.
+The idea is to represent variables without naming them.
+Instead, we can encode variables by their de Bruijn indexes, i.e. a reference to the binding abstraction represented as a natural.
+The natural number is the "nesting level" of the variable relative to its binding abstraction, starting at $0$.
+For example, $\lambda x. x$ can be encoded as $\lambda. 0,$ and $\lambda x\; y. x\;x\;y$ as $\lambda.\lambda. 1\;1\;0.$
+Two terms are alpha-equivalent if and only if their de Bruijn representation is the same.
+We should note that, during beta-reduction, de Bruijn indexes of free variables will change, but they will change predictably.
+Consider, for example, the following term:
+
+$\lambda. (\lambda. \lambda.\lambda. 1\;(2\;1\;0))\; 0$
+
+(this is $\lambda x.succ\;x$).
+It is quite apparent that $\lambda. \lambda.\lambda. 1\;(0\;1\;0)$ is not equivalent (injecting variable names back gives $\lambda x. \lambda s.\lambda z. s\;(z\;s\;z),$ when it should be $s\;(x\;s\;z)$).
+But, you might notice that $\lambda. \lambda.\lambda. 1\;(2\;1\;0)$ is (by the rule of eta-reduction).
+This line of reasoning gives us the intuition for the general rule.
+When an expression is substituted, de Bruijn indexes of its free variables must increase by the de Bruijn index of the variable the expression replaces.
+It should also be noted that when the left term contains free variables, those must be reduced by one, because application "strips" one level of $\lambda$, e.g.
+
+$\lambda. (\lambda. \lambda.\lambda. 1\;(2\;3\;0))\; 0$
+
+is $\lambda. \lambda.\lambda. 1\;(2\;2\;0),$
+not $\lambda. \lambda.\lambda. 1\;(2\;3\;0)$ (indeed in this example $3$ references a nonexistent abstraction).
+
+***
+
+</details>
+</p>
--- a/untyped-lambda-calculus/article.md
+++ b/untyped-lambda-calculus/article.md
@@ -159,10 +159,10 @@ Our decision points are:
 * to start with the outermost or the innermost redex;
 * to reduce redexes inside abstraction bodies or not.
 
-#### Direction of reducing
+#### Direction of reduction
 
 We can reduce an expression left-to-right or right-to-left.
-This decision is easy: going right-to-left is strictly less powerful because the process doesn't terminate on some terms whereas going left-to-right does.
+This decision is easy: going left-to-right is strictly more powerful because the process terminates on all terms for which going right-to-left does and then some. 
 
 #### Reducing outside-in or inside-out
 
@@ -181,6 +181,11 @@ Reducing the leftmost outermost redex first is called the _normal order strategy
 
 In other words, the applicative order strategy first reduces all of the arguments (from left to right) and then applies the abstraction, while normal order does the inverse.
 
+Expressions that are not reducible using the chosen evaluation strategy are said to be in the _normal form_.
+
+If a given term has a normal form under both of these strategies, it is the same for both.
+But the applicative order strategy is strictly weaker than the normal order strategy: some terms that normalize under normal order don't normalize under applicative order.
+
 #### Reducing abstraction bodies or not
 
 The two strategies above are free to go inside abstractions to find redexes.
@@ -190,11 +195,8 @@ But there are also strategies that can't go inside abstraction bodies, which we 
 When we're talking about programming languages, we can't necessarily inspect function bodies, so the weak strategies are considered more practical there.
 
 The weak strategy corresponding to the normal order is called _call-by-name_, and the one corresponding to the applicative order is _call-by-value_.
-Call-by-value corresponds to eager evaluation.
-Call-by-name, on the other hand, corresponds to lazy evaluation.
 
-In eager evaluation, arguments are always evaluated before the function call.
-Most programming languages follow this strategy by default.
+Call-by-name corresponds to lazy evaluation.
 In lazy evaluation, some arguments are potentially never evaluated at all, but some computations may be duplicated.
 Very few programming languages implement lazy evaluation this way (the only real example that comes to mind is Algol 60).
 
@@ -202,17 +204,27 @@ A more practical variation of call-by-name is _call-by-need_, which is, at least
 Call-by-need behaves exactly like call-by-name, except that reduction steps that would duplicate computations, don't.
 It does that by sharing the computation corresponding to an argument everywhere it appears, so any shared computation is performed at most once.
 
-To sum up, here's a table with all the mentioned strategies and their characteristics:
+Call-by-value, on the other hand, corresponds to eager evaluation.
+In eager evaluation, arguments are always evaluated before the function call.
+Most programming languages follow this strategy by default.
 
-| Name                 | Outermost- or innermost- first? | Strong or weak? |
-|----------------------|---------------------------------|-----------------|
-| Normal order         | Outermost                       | Strong          |
-| Applicative order    | Innermost                       | Strong          |
-| Call-by-name (eager) | Outermost                       | Weak            |
-| Call-by-value (lazy) | Innermost                       | Weak            |
-| Call-by-name         | Innermost                       | Weak            |
+To sum up, here's a table with the strategies mentioned and their characteristics:
 
-Expressions that are not reducible using the chosen evaluation strategy are said to be in the _normal form_.
+| Name                   | Outermost- or innermost- redex first? | Strong or weak? |
+|------------------------|---------------------------------------|-----------------|
+| Normal order           | Outermost                             | Strong          |
+| Applicative order      | Innermost                             | Strong          |
+| Call-by-name (lazy)    | Outermost                             | Weak            |
+| Call-by-value (eager)  | Innermost                             | Weak            |
+
+### Eta-reduction
+
+Speaking of reductions, let's also briefly mention _eta_-reduction ($\eta$-reduction).
+
+The rule of eta-reduction says that $\lambda x. f\;x = f$ if $x$ does not appear free in $f$.
+It boils down to the statement that two functions are the same if and only if they give the same result for all possible arguments.
+
+Eta-reduction is connected to the point-free style of writing Haskell since it enables to reduce functions like `id x = x` to `id`. 
 
 ## Programming in pure untyped lambda calculus
 
--- a/untyped-lambda-calculus/article.md
+++ b/untyped-lambda-calculus/article.md
@@ -2,33 +2,12 @@
 
 When talking about Haskell, the term "lambda calculus" often crops up.
 It is a theoretical framework used to define the meaning of computation in many functional languages.
-Understanding lambda calculus can be very helpful when talking about Haskell programs.
-It is also used extensively in programming language theory research and functional language design.
+Understanding lambda calculus can be very helpful when talking about functional programs, e.g. those written in Haskell, Agda, Idris, etc.
+Perhaps surprisingly, understanding lambda calculus can be also helpful in understanding C++ template metalanguage, because it's a functional language in its own right (although not, strictly speaking, based on lambda calculus).
+Lambda calculus is also used extensively in programming language theory research and functional language design.
 
 There are two broad kinds of lambda calculus: _untyped_ and _typed_.
 The _untyped_ kind, discussed in this article, expresses an unrestricted computation.
-However, as a system of logic, it is self-contradictory.
-The _typed_ kind, which we intend to discuss in a follow-up article, is a restricted version.
-It doesn't have the logical paradox that plagues the untyped kind and can serve as a logic system.
-
-The list of languages based on lambda calculus includes:
-
-- Haskell,
-- Most languages in the ML family, like OCaml and F#,
-- Most dependently typed languages, like Agda, Coq, F*, Idris,
-- Cardano blockchain's Plutus Core.
-
-There's a rumor that John McCarthy based Lisp on lambda calculus, but this seems to be, broadly speaking, false.
-The original Lisp used different name resolution semantics, and as much was stated by John himself:
-
-> ... one of the myths concerning LISP that people think up or invent for themselves becomes apparent, and that is that LISP is somehow a realization of the lambda calculus, or that was the intention.
-> The truth is that I didn't understand the lambda calculus, really.
->
-> -- John McCarthy, Lisp session, History of Programming Languages
-
-Also, while not strictly speaking based on lambda calculus, C++ template metalanguage is a functional language in its own right, so much of what we discuss below happens to be applicable.
-In fact, one could either embed lambda calculus into C++ template metalanguage or write a transpiler from a lambda calculus-based language into C++ template metalanguage.
-See, for instance, [Translating Lambda Calculus into C++ Templates by Vít Šefl](https://dx.doi.org/10.1007/978-3-030-83978-9_5).
 
 ## A bit of history
 
@@ -244,11 +223,11 @@ For example $and = \lambda x\;y. x\;y\;fls.$
 The $and$ function must return $tru$ if and only if both its arguments are $tru.$
 Let us make sure it does what we expect it to.
 
-$and\;tru\;tru = (\lambda x\;y. x\;y\;fls)\;(\lambda x\; y. x)\;(\lambda x\; y. x)$
+$and\;tru\;tru = (\lambda x\;y. x\;y\;fls)\;(\lambda x\; y. x)\;(\lambda x\; y. x).$
 
 We will rename the bound variables to avoid ambiguity:
 
-$(\lambda x\;y. x\;y\;fls)\;(\lambda a\; b. a)\;(\lambda c\; d. c)$
+$(\lambda x\;y. x\;y\;fls)\;(\lambda a\; b. a)\;(\lambda c\; d. c).$
 
 Now we can reduce the expression.
 Let's use the call-by-value strategy here.
@@ -413,7 +392,7 @@ $\to cdown\;((\lambda x. cdown\;(x\;x))\;(\lambda x. cdown\;(x\;x)))\;c_1$
 $\Rightarrow (\lambda x. cdown\;(x\;x))\;(\lambda x. cdown\;(x\;x))\;(pred\;c_1)$
 $\Rightarrow (\lambda x. cdown\;(x\;x))\;(\lambda x. cdown\;(x\;x))\;c_0$
 $\to cdown\;((\lambda x. cdown\;(x\;x))\;(\lambda x. cdown\;(x\;x)))\;c_0$
-$\Rightarrow c_0$
+$\Rightarrow c_0.$
 
 For illustrative purposes, we used full beta-reduction, carefully choosing which terms to reduce first.
 We could also have used call-by-name at the cost of intermediate expressions getting rather long-winded.
@@ -450,6 +429,13 @@ The last three points are somewhat profound if you think about it: we only defin
 There are other constructions of the universal computer: Turing and Post machines, the Markov algorithm, and general recursive functions.
 But, arguably, nothing beats the simplicity of lambda calculus.
 One has to appreciate the elegance at least.
+
+As for more practical applications, to name a few languages based on lambda calculus:
+
+- Haskell,
+- Most languages in the ML family, like OCaml and F#,
+- Most dependently typed languages, like Agda, Coq, F*, Idris,
+- Cardano blockchain's Plutus Core.
 
 ## Exercises
 
--- a/untyped-lambda-calculus/article.md
+++ b/untyped-lambda-calculus/article.md
@@ -1,15 +1,15 @@
 # Untyped Lambda Calculus
 
 When talking about Haskell, the term "lambda calculus" often crops up.
-It is a theoretical framework used to define the meaning of computation in many functional languages.
-Understanding lambda calculus can be very helpful when talking about functional programs, e.g. those written in Haskell, Agda, Idris, etc.
-Perhaps surprisingly, understanding lambda calculus can be also helpful in understanding C++ template metalanguage, because it's a functional language in its own right (although not, strictly speaking, based on lambda calculus).
+It's a theoretical framework used to define the meaning of computation in many functional languages.
+Understanding lambda calculus can be very helpful when talking about programs written in languages like Haskell, Agda, Idris, etc.
+
+Perhaps surprisingly, understanding lambda calculus can also be helpful in understanding the C++ template metalanguage since it's a functional language in its own right (although it's not, strictly speaking, based on lambda calculus).
 Lambda calculus is also used extensively in programming language theory research and functional language design.
 
 There are two broad kinds of lambda calculus: _untyped_ and _typed_.
-The _untyped_ kind, discussed in this article, expresses an unrestricted computation.
-
-While untyped lambda calculus doesn't find much use outside of computability theory, it is necessary to understand it to meaningfully discuss the typed variety (strictly speaking, varieties).
+The untyped kind, discussed in this article, expresses an unrestricted computation.
+While it doesn't find much use outside of computability theory, it's necessary to understand it to meaningfully discuss the typed varieties.
 
 ## A bit of history
 
@@ -46,15 +46,15 @@ Lambda calculus that only contains functions and variables is called pure, meani
 
 We can formally define the pure untyped lambda calculus in terms of _abstractions_ and _applications_.
 
-### Abstraction
+### Abstractions
 
 An _abstraction_, or a functional abstraction, is a parametric expression, that is to say, a function.
-Lambda calculus only defines univariate functions, but we can easily extend it to include multivariate functions too, as we'll see later.
+Lambda calculus only defines univariate (single-variable) functions, but we can easily extend it to include multivariate functions too, as we'll see later.
 
 We introduce new abstractions using the $\lambda$ symbol.
-The abstraction consists of the _head_ and the _body_, separated by the dot ($.$) symbol.
-The _head_ contains the $\lambda$ symbol and the parameter name.
-The _body_ is an arbitrary expression.
+An abstraction consists of a _head_ and a _body_, separated by the dot ($.$) symbol.
+The head contains the $\lambda$ symbol and the parameter name.
+The body is an arbitrary expression.
 For example, $\lambda x. x$ is an abstraction.
 
 As usual in abstract mathematics, we can assign a name to an expression.
@@ -74,7 +74,7 @@ The choice of bound variable names is arbitrary.
 Expressions that differ only in the names of bound variables are called _alpha-equivalent_.
 For example, $\lambda x. x$ and $\lambda y. y$ are alpha-equivalent, but $\lambda x. y$ and $\lambda y. z$ are not.
 
-### Application
+### Applications
 
 _Application_ is the operation of substituting a specific value for the parameter in the abstraction.
 It is the only operation defined in the pure untyped lambda calculus.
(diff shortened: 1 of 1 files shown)


untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,461 @@
...
+Lambda calculus, initially envisioned as a formal logic system, was developed by Alonzo Church around the 1930s to explore the foundations of mathematics.
+The initial formulation had a logical inconsistency known as [Kleene–Rosser paradox].
+To sidestep this issue, Church isolated the part of lambda calculus relevant only to computation in 1936.
+This isolate is now known as the untyped lambda calculus.
+Later, In 1940, a typed version of lambda calculus was introduced, which has weaker expressive power but is logically consistent.
+
+[Kleene–Rosser paradox]: https://en.wikipedia.org/wiki/Kleene%E2%80%93Rosser_paradox
+
+The lambda symbol used in the standard notation, from which lambda calculus derives its name, most likely was chosen arbitrarily.
+However, there's an apocryphal story the notation was derived from $\hat x$, morphing first into $\Lambda x$, and later into $\lambda x$.
```
Heimdell: Because publisher had no `^` litera, yes :D

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,461 @@
+# Untyped Lambda Calculus
+
+When talking about Haskell, the term "lambda calculus" often crops up.
+It is a theoretical framework used to define the meaning of "computation" in many functional languages.
+Understanding lambda calculus can be very helpful when talking about Haskell programs.
+It is also used extensively in programming language theory research and functional language design.
+
+There are two broad kinds of lambda calculus, _typed_, and _untyped_.
+The _untyped_ kind, discussed in this article, expresses an unrestricted computation.
+However, as a system of logic, it is self-contradictory.
```
NaeosPsy, line 10: Could you put a link that describes why? Perhaps here there's something: https://serokell.io/blog/formal-verification-history

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,461 @@
...
+
+The lambda symbol used in the standard notation, from which lambda calculus derives its name, most likely was chosen arbitrarily.
+However, there's an apocryphal story the notation was derived from $\hat x$, morphing first into $\Lambda x$, and later into $\lambda x$.
+
+In the mid-1960s, Peter Landin has shown that lambda calculus models arbitrarily complex programming languages.
+This insight arguably kickstarted the research on functional programming languages.
+
+## The pure untyped lambda calculus
+
+The untyped lambda calculus is, first and foremost, a consistent equational theory of functions.
```
NaeosPsy: Most of it is followable, but this sentence sticks out as some words that only those that know, know. Perhaps add a rephrasing or something?

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,461 @@
...
+There are two broad kinds of lambda calculus, _typed_, and _untyped_.
+The _untyped_ kind, discussed in this article, expresses an unrestricted computation.
+However, as a system of logic, it is self-contradictory.
+The _typed_ kind, which we intend to discuss in a follow-up article, is a restricted version.
+It doesn't have the logical paradox that plagues the untyped kind and can serve as a logic system.
+
+The list of languages based on lambda calculus includes:
+
+- Haskell
+- Most languages in the ML family, like OCaml and F#
```
NaeosPsy: Haskell is also in the ML family, no? Somewhat odd phrasing.

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,461 @@
+# Untyped Lambda Calculus
+
+When talking about Haskell, the term "lambda calculus" often crops up.
+It is a theoretical framework used to define the meaning of "computation" in many functional languages.
+Understanding lambda calculus can be very helpful when talking about Haskell programs.
+It is also used extensively in programming language theory research and functional language design.
+
+There are two broad kinds of lambda calculus, _typed_, and _untyped_.
```
NaeosPsy:
```suggestion
There are two broad kinds of lambda calculus: _untyped_ and _typed_.
```

untyped-lambda-calculus/article.md
DanielRrr: Haskell doesn't belong to the ML family

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,461 @@
...
+However, as a system of logic, it is self-contradictory.
+The _typed_ kind, which we intend to discuss in a follow-up article, is a restricted version.
+It doesn't have the logical paradox that plagues the untyped kind and can serve as a logic system.
+
+The list of languages based on lambda calculus includes:
+
+- Haskell
+- Most languages in the ML family, like OCaml and F#
+- Cardano blockchain's Plutus Core
+
```
DanielRrr: I would add dependently typed languages as well

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,461 @@
...
+>
+> -- James Iry, [A Brief, Incomplete, and Mostly Wrong History of Programming Languages][brief-incomplete-history]
+
+[brief-incomplete-history]: http://james-iry.blogspot.com/2009/05/brief-incomplete-and-mostly-wrong.html
+
+Lambda calculus, initially envisioned as a formal logic system, was developed by Alonzo Church around the 1930s to explore the foundations of mathematics.
+The initial formulation had a logical inconsistency known as [Kleene–Rosser paradox].
+To sidestep this issue, Church isolated the part of lambda calculus relevant only to computation in 1936.
+This isolate is now known as the untyped lambda calculus.
+Later, In 1940, a typed version of lambda calculus was introduced, which has weaker expressive power but is logically consistent.
```
DanielRrr: One may also mention that the idea of types has been initially proposed by Bertrand Russell (1903, 1907, 1911). The initial version of a typed lambda was a combination of the lambda calculus and Russell's type theory.

untyped-lambda-calculus/article.md
DanielRrr: I don't like the term 'equational theory' since it might be confusing with universal algebra. I would consider 'system of expression rewriting' or something like that.

DanielRrr (commented):
I'm gonna leave more comments a bit later

untyped-lambda-calculus/article.md
NaeosPsy: Fair enough. 😅

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,461 @@
...
+We call them strong because they go inside the abstraction bodies.
+The corresponding _weak_ evaluation strategies are called _call-by-name_ and _call-by-value_.
+Call-by-name is a weak version of the normal order strategy.
+Call-by-value is a little stricter and only allows reduction if the argument is a "value" -- that is, in pure lambda calculus, either a variable or an abstraction.
+
+It's interesting to note that the call-by-value strategy corresponds to eager evaluation.
+In eager evaluation, arguments are evaluated before the function call.
+Call-by-name, on the other hand, corresponds to lazy evaluation.
+In lazy evaluation, the arguments potentially are never evaluated at all.
+
```
DanielRrr: I think this part might require some examples to compare strategies and their pros and cons in the concrete case.

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,461 @@
...
+We run into a small issue, however.
+We don't have much to work with yet.
+We don't have booleans, numbers, tuples, or multivariate functions.
+No worries, though!
+We can represent all of that using only abstractions and applications!
+
+### Multivariate functions
+
+Any multivariate function is the same as multiple nested functions returning other functions.
+If you're familiar with functional programming, you might know such functions are called curried functions, named in honor of Haskell Curry.
```
DanielRrr: s/named in honor/named after

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,461 @@
...
+We will rename the bound variables to avoid ambiguity:
+
+$(\lambda x\;y. x\;y\;fls)\;(\lambda a\; b. a)\;(\lambda c\; d. c)$
+
+Now we can reduce the expression.
+Let's use call by value strategy here.
+The arrow $\to$ represents one evaluation step:
+
+$(\lambda x\;y. x\;y\;fls)\;(\lambda a\; b. a)\;(\lambda c\; d. c)$
+$\to (\lambda y. (\lambda a\; b. a)\;y\;fls)\;(\lambda c\; d. c)$
```
DanielRrr, line 297: I would use the $\to_{\beta}$ arrow

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,461 @@
...
+In particular, we can encode recursion using the fixed point combinator, also known as $Y$ combinator:
+
+$$Y = \lambda f. (\lambda x. f\;(x\;x))\;(\lambda x. f\;(x\;x)).$$
+
+A shorter equivalent form is sometimes cited:
+
+$$X = \lambda f. \Omega\; (\lambda x. f\;(x\;x)).$$
+
+We should note that it's not the only fixed-point combinator.
+For instance, another famous one is
```
DanielRrr: I would say explicitly that it's named after Turing

untyped-lambda-calculus/article.md
DanielRrr: And the Lisp family.

Perhaps, it's worth saying that mainstream languages mostly have lambdas as anonymous functions.

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,461 @@
...
+- Haskell
+- Most languages in the ML family, like OCaml and F#
+- Cardano blockchain's Plutus Core
+
+There's a rumor that John McCarthy based Lisp on lambda calculus, but this seems to be, broadly speaking, false.
+The original Lisp used different name resolution semantics, and as much was stated by John himself:
+
+> ... one of the myths concerning LISP that people think up or invent for themselves becomes apparent, and that is that LISP is somehow a realization of the lambda calculus, or that was the intention.
+> The truth is that I didn't understand the lambda calculus, really.
+>
```
DanielRrr: Don't believe him, lol

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,461 @@
...
+> 1936 - Alonzo Church also invents every language that will ever be but does it better.
+> His lambda calculus is ignored because it is insufficiently C-like.
+> This criticism occurs in spite of the fact that C has not yet been invented.
+>
+> -- James Iry, [A Brief, Incomplete, and Mostly Wrong History of Programming Languages][brief-incomplete-history]
+
+[brief-incomplete-history]: http://james-iry.blogspot.com/2009/05/brief-incomplete-and-mostly-wrong.html
+
+Lambda calculus, initially envisioned as a formal logic system, was developed by Alonzo Church around the 1930s to explore the foundations of mathematics.
+The initial formulation had a logical inconsistency known as [Kleene–Rosser paradox].
```
DanielRrr: It's worth referring to something like this https://plato.stanford.edu/entries/paradoxes-contemporary-logic/#IncoCertFormLogi

DanielRrr (commented):
Looks OK. I have several tiny remarks.

untyped-lambda-calculus/article.md
lierdakil: I don't think I use arrows in any other sense in this article, so I don't think noisier notation is justified here.

untyped-lambda-calculus/article.md
lierdakil: It's complicated with Lisp. The "wisdom of the crowds" insists McCarthy based Lisp on λ-calculus, but the man himself insists this was not the case. I'm inclined to believe the latter, because while it's obvious McCarthy was _aware_ of λ-calculus, AFAIK the first versions of Lisp didn't really fit into any λ-calculus evaluation strategy. Modern Lisp dialects are arguably similar enough to λ-calculus to say they're loosely based on it, but historically it seems to be more of a case of convergent evolution. It is pretty hard to untangle what inspired what, though.

lierdakil: As for "lambdas as anonymous functions", it's also kind of complicated. While the _syntax_ for anonymous functions is often loosely inspired by λ-calculus (often very indirectly), the actual semantics have little to do with it, so I don't know how useful it is to mention that.

untyped-lambda-calculus/article.md
lierdakil: As I outlined above, there is some corroborating evidence to his claim.

untyped-lambda-calculus/article.md
lierdakil: It's a little hard to construct examples without first introducing some meaningful terms, which is done below. I could of course use impure calculus (e.g. I could introduce naturals), but doing that for a one-off example seems kind of extraneous.

I'll see what I can do with simple terms, but I'm not very optimistic frankly.

untyped-lambda-calculus/article.md
lierdakil: I've added a mention that typed λ-calculus is based on Russel's type theory. Getting into it beyond that seems slightly off-topic. Hopefully I'll remember to get into more detail in the planned simply typed λ-calculus article.

untyped-lambda-calculus/article.md
lierdakil: Well, I've added _an_ example. It illustrates some usual points, but I can't say I'm very happy with how it turned out. I expect most readers will skip right over.

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,493 @@
...
+
+There's a rumor that John McCarthy based Lisp on lambda calculus, but this seems to be, broadly speaking, false.
+The original Lisp used different name resolution semantics, and as much was stated by John himself:
+
+> ... one of the myths concerning LISP that people think up or invent for themselves becomes apparent, and that is that LISP is somehow a realization of the lambda calculus, or that was the intention.
+> The truth is that I didn't understand the lambda calculus, really.
+>
+> -- John McCarthy, Lisp session, History of Programming Languages
+
+Also, while not strictly speaking based on lambda calculus, C++ template metalanguage is a functional language in its own right, so much of what we discuss below happens to be applicable.
```
NaeosPsy: Is this necessary? I feel like we can cut this.

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,493 @@
...
+
+The lambda symbol used in the standard notation, from which lambda calculus derives its name, most likely was chosen arbitrarily.
+However, there's an apocryphal story the notation was derived from $\hat x$, morphing first into $\Lambda x$, and later into $\lambda x$.
+
+In the mid-1960s, Peter Landin has shown that lambda calculus models arbitrarily complex programming languages.
+This insight arguably kickstarted the research on functional programming languages.
+
+## The pure untyped lambda calculus
+
+The untyped lambda calculus is, first and foremost, a consistent expression rewriting system.
```
NaeosPsy: Would "consistent system for rewriting expressions" also be correct? I feel it would sound better.

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,493 @@
...
+Lambda calculus, initially envisioned as a formal logic system, was developed by Alonzo Church around the 1930s to explore the foundations of mathematics.
+The initial formulation had a logical inconsistency known as [Kleene–Rosser paradox] (for a more comprehensive overview, refer to, e.g. [Paradoxes and Contemporary Logic][paradoxes-contemporary-logic]).
+To sidestep this issue, Church isolated the part of lambda calculus relevant only to computation in 1936.
+This isolate is now known as the untyped lambda calculus.
+Later, In 1940, a typed version of lambda calculus, based on Russel's type theory, was introduced, which has weaker expressive power but is logically consistent.
+
+[Kleene–Rosser paradox]: https://en.wikipedia.org/wiki/Kleene%E2%80%93Rosser_paradox
+[paradoxes-contemporary-logic]: https://plato.stanford.edu/entries/paradoxes-contemporary-logic/#IncoCertFormLogi
+
+The lambda symbol used in the standard notation, from which lambda calculus derives its name, most likely was chosen arbitrarily.
```
NaeosPsy: This is a cool fact. But it feels that right here, a person that has not seen the syntax at all (since you introduce it later) will not find it relatable. Perhaps cut it or move it below?

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,493 @@
...
+However, there's an apocryphal story the notation was derived from $\hat x$, morphing first into $\Lambda x$, and later into $\lambda x$.
+
+In the mid-1960s, Peter Landin has shown that lambda calculus models arbitrarily complex programming languages.
+This insight arguably kickstarted the research on functional programming languages.
+
+## The pure untyped lambda calculus
+
```
NaeosPsy, line 41: It feels that it would be nice to start with some justification for "pure" just so it flows a little bit better

E.g. something like this: 

The best example for starters is pure untyped lambda calculus, "pure" meaning that it only contains functions and variables.

This calculus is, first and foremost, a consistent system for rewriting expressions.

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,493 @@
...
+The untyped lambda calculus is, first and foremost, a consistent expression rewriting system.
+Calculus that only contains functions and variables is called the pure untyped lambda calculus, "pure" meaning it doesn't have anything beyond what's strictly necessary.
+
+We can formally define the pure untyped lambda calculus in terms of _abstractions_ and _applications_.
+
+_Abstraction_, or functional abstraction, is a definition of a parametric expression, that is to say, a function.
+Lambda calculus only defines univariate functions, but that's easily extended to multivariate, as we'll see later.
```
NaeosPsy: I find "that's easily extended to multivariate" a bit confusing even though I understand what you mean.

Perhaps "it can be easily extended" as in lambda calculus, or those as in functions? Or some other rewording.

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,493 @@
...
+This insight arguably kickstarted the research on functional programming languages.
+
+## The pure untyped lambda calculus
+
+The untyped lambda calculus is, first and foremost, a consistent expression rewriting system.
+Calculus that only contains functions and variables is called the pure untyped lambda calculus, "pure" meaning it doesn't have anything beyond what's strictly necessary.
+
+We can formally define the pure untyped lambda calculus in terms of _abstractions_ and _applications_.
+
+_Abstraction_, or functional abstraction, is a definition of a parametric expression, that is to say, a function.
```
NaeosPsy: Is abstraction an action or an item? 

If it is an item, then it should be _an abstraction_ I think. If it is the action like application, then it seems that it would be the _the act of defining a parametric expression_. I think that would be clearer. 

Above as well, perhaps it should be "in terms of two actions: abstraction and application".

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,493 @@
...
+
+You might notice that using whitespace can lead to ambiguity.
+We'll use parentheses where necessary to avoid it.
+Additionally, we will assume that application is left-associative and binds more tightly than abstractions.
+Hence, $f\;a\;b\;c$ is $((f\;a)\;b)\;c$, i.e. $f$ is applied to $a$, then the result is applied to $b$, then to $c$.
+Since application binds more tightly, we don't need parentheses around the body of an abstraction.
+However, we might need parenthesis around the whole abstraction itself.
+We will also eschew parentheses for nested abstractions, e.g. $\lambda x. \lambda y. x\;y$ is $\lambda x. (\lambda y. (x\;y)).$
+
+Variables in the body of an abstraction that are parameters of said abstraction are called _bound variables_.
```
NaeosPsy: Feels like free & bound variables & alpha-equivalence go with "abstraction", so they should probably be moved up? 

I believe you could make level-3 subheaders for abstraction (where is everything regarding variables) and application, it would look clearer like that.

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,493 @@
...
+Other variables are called _free variables_.
+
+An expression without free variables is called a _closed term_ or a _combinator_.
+
+The choice of names of bound variables is arbitrary.
+Expressions that differ only in the names of bound variables are called _alpha-equivalent_.
+For example, $\lambda x. x$ and $\lambda y. y$ are alpha-equivalent, but $\lambda x. y$ and $\lambda y. z$ are not.
+
+## Defining the computation
```
NaeosPsy:
```suggestion
## Defining computation in lambda calculus
```

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,493 @@
...
+
+The step of the computation is called beta-reduction.
```
NaeosPsy: This sentence I don't really like. 

Perhaps something like: A step of computation in lambda calculus is called beta reduction.

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,493 @@
...
+
+For example, the expression $(\lambda x. x) y$ is a redex, and one step of beta-reduction transforms it to just $y$.
+
+Per rule 1, we rename bound variables if there might be ambiguity.
+For example, $(\lambda x. \lambda y. x\;y)\;(\lambda x. x\;y)$ is a redex.
+However, the $y$ identifier is ambiguous.
+In the left term, it refers to the bound variable; in the right one, it refers to a free variable.
+To avoid ambiguity, we can rename $y$ in the first expression to, say, $z$: $(\lambda x. \lambda z. x\;z)\;(\lambda x. x\;y).$
+Then we can do the rest of the steps with no fear of confusion: $\lambda z. (\lambda x. x\;y)\;z.$
+
```
NaeosPsy, line 141: Feel like this should be a subsection called evaluation strategies.

untyped-lambda-calculus/article.md
NaeosPsy: Yeah, I don't think it is really helpful. 

If you use one in the article, I would introduce how that works, tell that there are options, perhaps a fun fact or two about Haskell, but wouldn't explain this all, it doesn't really help the reader.

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,493 @@
...
+### Church booleans
+
+Let us now introduce booleans.
+We will follow Church here.
+However, it's not the only way.
+Arguably there are infinitely many ways to encode values as functions.
+
+So, let us declare two combinators,
+$tru = \lambda x\; y. x,$ and $fls = \lambda x\;y. y.$
+We're calling these $tru$ and $fls$ to disambiguate them from atomic boolean values.
```
NaeosPsy: What is an atomic boolean value?

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,493 @@
...
+
+To simplify the notation, we will now introduce a shorthand syntax.
+A multivariate abstraction of $n$ parameters, declared as $\lambda x_1\;x_2\;\ldots\;x_n. t,$ is the same as $n$ nested abstractions $\lambda x_1. \lambda x_2. \ldots \lambda x_n. t.$
+
+Multivariate abstractions, naturally, can be partially applied, e.g. $(\lambda x\; y\; z. (x\;y)\;(x\;z))\;a\;b$ reduces to $\lambda z. (a\;b)\;(a\;z).$
+
+### Church booleans
+
+Let us now introduce booleans.
+We will follow Church here.
```
NaeosPsy: Do you think it is clear what this sentence means? Like, I understood, but some might be confused that you mean that you will cover the Church encoding of booleans

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,493 @@
...
+
+That also works.
+The remaining two cases are left as an exercise for the reader, but it should be rather obvious that the combinator works as intended.
+Indeed, since $tru$ essentially chooses its first argument and $fls$ its second, $x\;y\;fls$ essentially means "if $x$ is $tru$, the result is $y$, otherwise $fls$".
+
+Defining other boolean combinators is reasonably straightforward by analogy.
+
+### Church pairs
+
+Since we have booleans now, we can also encode pairs (and by extension, tuples as trees or right combs) as functions, returning one value when given $tru$ as an argument, and another when given $fls$.
```
NaeosPsy: Not sure it is worth it to mention about tuples as trees and right combs.

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,493 @@
...
+
+$fst\;(pair\;x\;y)$
+$= (\lambda p. p\;tru)\;((\lambda f\;s\;b. b\;f\;s)\;x\;y)$
+$\Rightarrow (\lambda p. p\;tru)\;(\lambda b. b\;x\;y)$
+$\to (\lambda b. b\;x\;y)\;tru$
+$\to tru\;x\;y$
+$= (\lambda a\; b. a)\;x\;y$
+$\Rightarrow x.$
+
+Here, double arrow $\Rightarrow$ signifies more than one step of beta-reduction.
```
NaeosPsy: more than one step -> multiple steps of beta-reduction?

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,493 @@
...
+
+### Church naturals
+
+Since we encoded both pairs and booleans, we already can encode naturals simply as sequences of booleans, that is, as binary numbers.
+In practice, this turns out to be cumbersome, however, so instead, we'll define naturals similar to the Peano construction.
```
NaeosPsy: Can you add some link to Peano numbers for the teapots in the audience, perhaps?

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,493 @@
...
+
+Now we can define a successor function, taking a natural and returning the next natural.
+Since we defined naturals as functions of two arguments, we expect the successor function to be a function of three arguments (i.e. three nested abstractions).
+The body should apply the "next" argument to the natural one more time:
+
+$succ = \lambda n\; s\; z. s\;(n\;s\;z).$
+
+It might be easier to see what's going on if we explicitly note the outer abstraction:
+
+$succ = \lambda n. \lambda s\; z. s\;(n\;s\;z).$
```
NaeosPsy, line 374: You use both $..$ and $$...$$, is that on purpose? 

It seems that $..$ renders much better for a text that uses a lot of math since $$...$$ gets centered which breaks the flow of the text too much. Perhaps we can just write everything with $..$?

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,493 @@
...
+
+The idea here is to give us the ability to express recursion by passing the function as its first argument.
+The omega combinator infinitely replicates itself.
+The $Y$ combinator exploits the same potentially infinitely replicating structure to encode recursion.
+
+Let us see how it works in practice.
+Consider a function that counts down to zero using the $pred$ combinator, defined in the previous section, and returns zero once the argument is zero.
+To define it, we will add a recursive call as its first argument:
+
+$cdown = \lambda f. \lambda n. ifThenElse\;(isZero\;n)\;c_0\;(f\;(pred\;n)).$
```
NaeosPsy, line 443: Point at the end is the ending of the sentence? 

Since it renders in a new line, I don't feel it is necessary here and elsewhere.

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,493 @@
...
+$$mul = \lambda m\; n. \lambda s\; z. m\;(n\;s)\;z.$$
+The former counts $m$ starting from $n$, and the latter counts $m$ times $n$ starting from $z.$
+
+We can also easily test whether a value corresponds to zero (i.e. $c_0$) or not:
+$$isZero = \lambda n. n\;(\lambda x. fls)\;tru.$$
+
+We run into a bit of an issue when trying to define subtraction or a predecessor function.
+Turns out it's doable but neither pretty nor efficient.
+We'll use pairs of values containing two subsequent naturals.
+That way, we can stagger the counter by one.
```
NaeosPsy: Is  "stagger" specific terminology? I don't really understand the use of this word here.

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,493 @@
...
+$ss = \lambda nn. pair\;(snd\; nn)\;(succ\;(snd\;nn)).$\
+
+The first element of the pair here is the predecessor, and the second is the successor.
+$zz$ encodes the starting point, and $ss$ is the staggered successor.
+
+Now we can define the predecessor function:
+
+$$pred = \lambda n. fst\; (n\; ss\; zz).$$
+
+With this definition, $pred\;c_0 = c_0,$ which isn't ideal.
```
NaeosPsy, line 404: I would put the comma outside the latex, It renders two different commas, which isn't very aesthetic, I think.

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,493 @@
...
+3. Define an exponentiation function for Church-encoded naturals.
+
+4. Define subtraction of Church-encoded naturals using the predecessor function `pred`.
+    For the case when the difference is undefined in naturals, i.e. when the subtrahend is bigger than minuend, feel free to return $c_0$ or $\Omega\;\Omega$.
+
+5. Define a function taking two Church-encoded naturals and returning `tru` if those are equal and `fls` otherwise.
+
+6. Implement a pure untyped lambda calculus interpreter in your favorite programming language.
+
+    One question you might have is how to handle name collisions and encode alpha-equivalence.
```
NaeosPsy, line 525: I would hide this under some Tips for handling name collisions spoiler or something, it puts too much text between the end of "real text" and more articles + subscriber box, which isn't aesthetic.

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,493 @@
+# Untyped Lambda Calculus
+
+When talking about Haskell, the term "lambda calculus" often crops up.
+It is a theoretical framework used to define the meaning of "computation" in many functional languages.
+Understanding lambda calculus can be very helpful when talking about Haskell programs.
+It is also used extensively in programming language theory research and functional language design.
+
+There are two broad kinds of lambda calculus: _untyped_ and _typed_.
```
NaeosPsy: The introduction feels a bit too confused and a tad bit too long. 

Perhaps you want to just mention that it is at the bottom of many functional programming languages and used extensively, and add a "Where is lambda calculus used?" section at the end?

I think this would improve readability (and SEO as well 😅).

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,505 @@
...
+For example, $\lambda x. x$ and $\lambda y. y$ are alpha-equivalent, but $\lambda x. y$ and $\lambda y. z$ are not.
+
+## Defining the computation
+
+The step of the computation is called beta-reduction.
+It is a single application of one expression to another.
+Formally, beta-reduction uses the following rules.
+Given an expression of the form $(\lambda x. t_1)\; t_2$ (i.e. an application of some abstraction to some expression $t_2$):
+
+1. Rename bound variables in $t_1$ and $t_2$ to avoid ambiguity.
```
Heimdell: Maybe, introduce de Bruijn indices ~and alpha-equivalence here~?

untyped-lambda-calculus/article.md
Heimdell: No, that's "we don't use de Bruijn indices" section.

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,505 @@
...
+$\to (\lambda z. ((\lambda z. z)\; a)\; ((\lambda z. z)\; a)\; ((\lambda w. w)\; b))\; ((\lambda p. p)\; c)$ \
+$\to ((\lambda z. z)\; a)\; ((\lambda z. z)\; a)\; ((\lambda w. w)\; b)$ \
+$\to a\; ((\lambda z. z)\; a)\; ((\lambda w. w)\; b)$ \
+$\to a\; a\; ((\lambda w. w)\; b)$ \
+$\to a\; a\; b$
+
+Note how in call-by-name we didn't have to compute $((\lambda p. p)\; c)$ argument, because it disappeared when applying.
+On the other hand, we had to compute $((\lambda z. z)\; a)$ twice.
+
+A practically significant variation on call-by-name is _call-by-need_, which is, at least in theory, used in Haskell.
```
Heimdell: I'd call it "reduce outermost redex with sharing" strategy

untyped-lambda-calculus/article.md
Heimdell: Or https://wiki.c2.com/?ChurchOfTheLeastFixedPoint.

untyped-lambda-calculus/article.md
Heimdell: `true` and `false` in "normal" languages?

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,505 @@
...
+
+The first element of the pair here is the predecessor, and the second is the successor.
+$zz$ encodes the starting point, and $ss$ is the staggered successor.
+
+Now we can define the predecessor function:
+
+$$pred = \lambda n. fst\; (n\; ss\; zz).$$
+
+With this definition, $pred\;c_0 = c_0,$ which isn't ideal.
+However, we don't have an encoding for negative numbers, so we don't have much of a choice.
```
Heimdell, line 405: Strictly speaking, we can _make_ one.

Heimdell (approved):
Very good!

untyped-lambda-calculus/article.md
lierdakil: While not strictly speaking _necessary_ (nothing in the intro is strictly speaking necessary besides "λ-calculus is every FP programming language invented, ever" :laughing: ), I, for one, find this curious, and it suggests that lambda-calculus is applicable outside the hard-core FP bubble. If the point of the intro is to provide the motivation for the reader, I think this paragraph is useful. I'd rather trim down the LISP discussion if you think the intro is a bit too long.

NaeosPsy: I see. In that case, perhaps we can just follow this comment: https://github.com/serokell/blog-posts/pull/12#discussion_r857395866

untyped-lambda-calculus/article.md
NaeosPsy:
```suggestion
We're calling these $tru$ and $fls$ to disambiguate them from "regular" boolean values, which are atomic.
```

Something like this, perhaps.  It is clear what is meant from the surrounding context, but it can raise an unnecessary "huh", I think.

untyped-lambda-calculus/article.md
lierdakil: > start with some justification for "pure"

To me, this feels weird. Specifically, why are we explaining what is "pure" without explaining what λ-calculus is? "Pure" is a qualifier on λ-calculus. It's like explaining a "pink flamingo" starting from "pink". It's a flamingo that happens to be pink, why are we starting with the least important bit? This feels backwards.

untyped-lambda-calculus/article.md
lierdakil: Technically, it's the same words in slightly different order, so it's not wrong. Can't say I like it, but that's why we have an editor, I guess :laughing:

Fixed.

untyped-lambda-calculus/article.md
lierdakil: Good point, moved.

untyped-lambda-calculus/article.md
lierdakil: Reworded to

> Lambda calculus only defines univariate functions, but we can easily extend it to include multivariate functions too, as we'll see later.

untyped-lambda-calculus/article.md
lierdakil: Uh. No, abstraction is a term, it doesn't act on terms, that's a confusing word choice on my part. Rephrased to

> An _abstraction_, or a functional abstraction, is a parametric expression, that is to say, a function.

untyped-lambda-calculus/article.md
lierdakil: Good point, moved.

Although I'm not sure about the sub-headers. FWIW I've added them, but it feels kind of meh.

untyped-lambda-calculus/article.md
lierdakil: Ehh... I'm not sure about this change. See, the header is not wrong, we're actually defining _the_ computation by the merit of the fact that every possible computation is representable in terms of λ-calculus :confused: We could define it any other way (Turing machines, Post machines, μ-recursive functions, Markov's normalized algorithms, etc), but it'll still be the same thing, all definitions are mathematically equivalent.

untyped-lambda-calculus/article.md
lierdakil: Now this is fine. Changed.

untyped-lambda-calculus/article.md
lierdakil: "Atomic" is non-decomposable, i.e. it doesn't have any internal structure (hence very different from Church booleans that are actually abstractions). It's a pretty common expression in math circles, but I see your point, a casual reader might not be familiar. Tried to explain it better:

> We're calling these $tru$ and $fls$ to disambiguate them from actual boolean values $true$ and $false$ that aren't actually abstractions.
> $true$ and $false$ are, hopefully obviously, not a part of the pure lambda calculus, but we can introduce them as terms in an impure lambda calculus.

untyped-lambda-calculus/article.md
NaeosPsy: I believe the second would still agree with what you say. 

For me, saying "defining the computation" sounds wrong on a feels level. Could be because computation is a process, and those don't take "the"?

untyped-lambda-calculus/article.md
NaeosPsy: Yeah, I understood it once I thought about it. Sounds good, thank you. 👍

untyped-lambda-calculus/article.md
lierdakil: @NaeosPsy I've added a level 3 header where it seemed appropriate (i.e. after "It brings us to the discussion of evaluation strategies."). The highlighted part is not it, though.

@Heimdell :unamused: I actually introduce de Bruijn encoding in the exercises section. It's nice when implementing λ-calculus in software, but not so well suited for napkin math we're doing in this here article.

BTW, you might want to check my explanation of de Bruijn encoding, I'm not sure if it's quite on-point because it's been a while since I implemented my λ-calculus evaluation engine.

untyped-lambda-calculus/article.md
lierdakil: Changed to

> We will do it in the same way Church did in his original paper.

Although I was briefly tempted to add "the" before "Church" and call it a day :joy:

untyped-lambda-calculus/article.md
lierdakil: I can drop "as trees or right combs", that's a reasonably extraneous remark.

untyped-lambda-calculus/article.md
lierdakil: Technically, "multiple" is ill-defined, as it can depending on the context mean "one or more" or "more than one", but in this case it doesn't really matter. Changed.

untyped-lambda-calculus/article.md
NaeosPsy: Yeah, ok, makes sense. 

The main problem here was that there is no justification for introducing exactly "pure untyped lambda calculus"; there's only definitions. 

So you can also have the same sentence order as before, just I'd suggest to add something like: "the most basic version of this calculus is pure untyped lambda calculus, which contains "

untyped-lambda-calculus/article.md
lierdakil: Yep. Can't decide between [Wikipedia](https://en.wikipedia.org/wiki/Peano_axioms#Historical_second-order_formulation) and [Haskell Wiki](https://wiki.haskell.org/Peano_numbers). Haskell wiki entry is short and simple, but relies on the knowledge of Haskell. The Wikipedia article of course goes into much more (arguably, here, unnecessary) detail.

untyped-lambda-calculus/article.md
lierdakil: Here, we can, yes.

untyped-lambda-calculus/article.md
lierdakil: Okay, I may be misusing the word here, but I definitely encountered "staggered counter" in the literature (meaning a counter lagging by some specific offset). Rewrote to

> That way, we can have a counter that's lagging by one.

untyped-lambda-calculus/article.md
NaeosPsy: I think Haskell Wiki is fine, Wikipedia one looks too scary for anybody who would need that link.

untyped-lambda-calculus/article.md
NaeosPsy: Oh, ok. Yes, that's definitely clearer.

untyped-lambda-calculus/article.md
lierdakil: The reason I put commas and other punctuation marks inside LaTeX is that otherwise those punctuation marks will eventually run away to the next line. Arguably, this is worse than a slight font mismatch:

image

untyped-lambda-calculus/article.md
NaeosPsy: You used the word later as well, btw, that should also probably be changed. (On line 361 for me)

untyped-lambda-calculus/article.md
NaeosPsy: It is indeed worse. Okay. 😁

untyped-lambda-calculus/article.md
lierdakil: I found no hard evidence for this, though, so this might be folklore.

untyped-lambda-calculus/article.md
lierdakil: As I said above, de Bruijn indices aren't very helpful when doing napkin math. I do introduce them when they become relevant (i.e. in exercises)

untyped-lambda-calculus/article.md
lierdakil: Oh, strictly speaking, we can make anything computable. The point is to define subtraction given this encoding.

untyped-lambda-calculus/article.md
lierdakil: Eh, that's actually standard practice for formulae-heavy articles. It's harder to parse without puncutation.

NaeosPsy: Ok, if that's the case, then I have no objections.

untyped-lambda-calculus/article.md
lierdakil: Well, here's the slightly updated version:

> The simplest, smallest type of lambda calculus is the pure untyped lambda calculus, so we'll start with it.

> The untyped lambda calculus is, first and foremost, a consistent system for rewriting expressions.
> Lambda calculus that only contains functions and variables is called pure, meaning it doesn't have anything beyond what's strictly necessary.

untyped-lambda-calculus/article.md
lierdakil: I've split the difference and removed "the". Feel free to argue if you feel that "in lambda calculus" is important (but IMO it's implied anyway, and shorter, catchier titles = more better in my book)

untyped-lambda-calculus/article.md
lierdakil: I've cut down the intro a bit (arguably a lot), and moved the list of languages based on λ-calculus to the conclusion. Hopefully this is will work better for SEO :laughing:

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,489 @@
+# Untyped Lambda Calculus
+
+When talking about Haskell, the term "lambda calculus" often crops up.
+It is a theoretical framework used to define the meaning of computation in many functional languages.
+Understanding lambda calculus can be very helpful when talking about functional programs, e.g. those written in Haskell, Agda, Idris, etc.
+Perhaps surprisingly, understanding lambda calculus can be also helpful in understanding C++ template metalanguage, because it's a functional language in its own right (although not, strictly speaking, based on lambda calculus).
+Lambda calculus is also used extensively in programming language theory research and functional language design.
+
+There are two broad kinds of lambda calculus: _untyped_ and _typed_.
+The _untyped_ kind, discussed in this article, expresses an unrestricted computation.
```
NaeosPsy: This needs one or two more sentences, probably one with information like this: 

"the untyped kind doesn't find much use beyond the computability theory.
Still, it's necessary to understand the untyped lambda calculus to meaningfully discuss the typed variety (or, more accurately, varieties)."

lierdakil: Done.

untyped-lambda-calculus/article.md
NaeosPsy: This doesn't really sound good. 😅

Would you be fine if we switched the last two sentences to: 

Like any untyped lambda calculus, it is first and foremost a consistent system for rewriting expressions.
And it is called pure because it doesn't have anything beyond what's strictly necessary – it only contains functions and variables.

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,492 @@
...
+Here we'll be using the equals sign ($=$) for that.
+For example: $id = \lambda x. x.$
+
+Lambda calculus derives its name from the $\lambda$ symbol used in this notation.
+The symbol was most likely chosen arbitrarily.
+However, there's an apocryphal story the notation was derived from $\hat x$ morphing first into $\Lambda x$ and later into $\lambda x$.
+
+Variables in the body of an abstraction that are parameters of said abstraction are called _bound variables_.
+Other variables are called _free variables_.
+
```
NaeosPsy, line 66: Can we have some examples here? Since these terms are used quite often, I would make sure they are nailed down.

untyped-lambda-calculus/article.md
```diff
@@ -148,19 +162,6 @@ A practically significant variation on call-by-name is _call-by-need_, which is,
...
-
-In theory, Haskell strives to implement call-by-need.
-In practice, it is a little more complicated.
-The reason is simple -- the optimizer can make non-obvious transformations, which could either remove some duplicated computations or vice versa.
-
-***
-
-</details>
-</p>
-
```
lierdakil: @NaeosPsy, ?

Did you mean to move it somewhere, or did you decide this is irrelevant?

untyped-lambda-calculus/article.md
lierdakil: Fine by me. Done.

untyped-lambda-calculus/article.md
lierdakil: Added a couple examples. Sorry if the phrasing is a bit awkward, I'm feeling a little under the weather today, so words come a little harder than usual.

untyped-lambda-calculus/article.md
NaeosPsy: I did decide it is not very relevant. It also looked very ugly when rendered so I thought it would be better if it was gone.

lierdakil: Well, it's arguably an important caveat to keep in mind, but I guess it's fine to remove, as it is only tangentially related to the topic at hand. :ok_hand:

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,508 @@
...
+4. Define subtraction of Church-encoded naturals using the predecessor function `pred`.
+    For the case when the difference is undefined in naturals, i.e. when the subtrahend is bigger than minuend, feel free to return $c_0$ or $\Omega\;\Omega$.
+
+5. Define a function that takes two Church-encoded naturals and returns `tru` if those are equal and `fls` otherwise.
+
+6. Implement a pure untyped lambda calculus interpreter in your favorite programming language.
+
+<p>
+<details>
+<summary>Tips on handling name collisions</summary>
```
NaeosPsy: Unfortunately, this doesn't go with the list, I assume there needs to be some indentation. 🤔

image

NaeosPsy:
```
4. As you might remember, a semigroup is a monoid without the identity element. Can you come up with an example of a structure that is a semigroup but is not a monoid? 

    <details>
      <summary>Solution</summary>
    
	One example is [non-empty lists](https://hackage.haskell.org/package/base-4.16.0.0/docs/Data-List-NonEmpty.html) with the `++` operator. While a regular list can be empty, and the empty list thus satisfies the requirement of an identity element, non-empty lists have no such element.
	
    <hr>
    </details>
```

Here's what worked for my monoid article, if that helps.

NaeosPsy:
Did the second read, all looks good for me except the evaluation strategy section which still feels a bit too much/complicated/hard-to-read.

untyped-lambda-calculus/article.md
```diff
@@ -124,12 +126,12 @@ Any expression where beta-reduction can be applied, we'll call a redex (short fo
...
 For example, the expression $(\lambda x. x) y$ is a redex, and one step of beta-reduction transforms it to just $y$.
 
-Per rule 1, we rename bound variables if there might be ambiguity.
+According to the rule, we rename bound variables if there might be ambiguity.
 For example, $(\lambda x. \lambda y. x\;y)\;(\lambda x. x\;y)$ is a redex.
 However, the $y$ identifier is ambiguous.
 In the left term, it refers to the bound variable; in the right one, it refers to a free variable.
 
-To avoid ambiguity, we can rename $y$ in the first expression to, say, $z$:
+To avoid ambiguity, we can rename $y$ in the first expression to, say, $z$:
```
lierdakil: eh, the comment about spurious space was meant to go here.

untyped-lambda-calculus/article.md
lierdakil: My renderer renders `details` in a list item ugly, but maybe it's not an issue on the blog itself. Pushed a fixup.

lierdakil:
> the evaluation strategy section which still feels a bit too much/complicated/hard-to-read.

I'll see if I can do something about it later today.

lierdakil:
I've tried rewriting the evaluation strategies section. Don't know if that helped at all, but at least I've caught an embarrassing mistake :sweat_smile: (TL;DR: I've called all strategies normalizing, due to a combination of bad phrasing and an unfortunate edit; this is demonstrably false; I'm a bit surprised nobody seemed to notice)

untyped-lambda-calculus/article.md
```diff
@@ -146,63 +146,73 @@ This brings us to the discussion of evaluation strategies.
...
+* to start with the left or right side of the expression;
+* to start with the outermost or the innermost redex;
+* to reduce redexes inside abstraction bodies or not.
 
-Second, we'll decide whether we're first reducing the outermost or the innermost redex.
-To see what is meant by outermost and innermost, consider the following expression:
+#### Direction of reducing
+
+We can reduce an expression left-to-right or right-to-left.
+This decision is easy: going right-to-left is strictly less powerful because the process doesn't terminate on some terms whereas going left-to-right does.
```
lierdakil: Okay, my fault, but this doesn't sound right. What this meant to say is probably

> going right-to-left is strictly less powerful because the process doesn't terminate on all terms going left-to-right does, but not vice versa.

untyped-lambda-calculus/article.md
```diff
@@ -146,63 +146,73 @@ This brings us to the discussion of evaluation strategies.
...
-First, we need to decide whether we're going left-to-right or right-to-left.
-This decision happens to be easy: going right-to-left is strictly less powerful, i.e. the process doesn't terminate on some terms whereas going left-to-right does.
+Our decision points are: 
+* to start with the left or right side of the expression;
+* to start with the outermost or the innermost redex;
+* to reduce redexes inside abstraction bodies or not.
 
-Second, we'll decide whether we're first reducing the outermost or the innermost redex.
-To see what is meant by outermost and innermost, consider the following expression:
+#### Direction of reducing
```
lierdakil: Sounds awkward. Perhaps 

> Direction of reduction

?

untyped-lambda-calculus/article.md
```diff
@@ -146,63 +146,73 @@ This brings us to the discussion of evaluation strategies.
...
-Note that applicative order strategy is strictly weaker than normal order strategy.
-Some terms that normalize under normal order don't under applicative order.
-Likewise for the weak strategies.
+| Name                 | Outermost- or innermost- first? | Strong or weak? |
+|----------------------|---------------------------------|-----------------|
+| Normal order         | Outermost                       | Strong          |
+| Applicative order    | Innermost                       | Strong          |
+| Call-by-name (eager) | Outermost                       | Weak            |
+| Call-by-value (lazy) | Innermost                       | Weak            |
+| Call-by-name         | Innermost                       | Weak            |
```
lierdakil: Call-by-name is mentioned twice. I think you meant call-by-need the second time?

A little strange having it in this table here, as it is entirely unclear what differentiates it from call-by-name. It is also considered "lazy evaluation". Actually, under some definition of "lazy", call-by-need is the laziest of them all.

untyped-lambda-calculus/article.md
```diff
@@ -146,63 +146,73 @@ This brings us to the discussion of evaluation strategies.
...
 
-Furthermore, if a given term has a normal form at all, then reduction under the normal order strategy will always reach it.
-We will discuss terms that do not have a normal form a little later.
-
-Speaking of reductions, let's also mention _eta_-reduction ($\eta$-reduction).
-The rule of eta-reduction says that $\lambda x. f\;x = f,$ if $x$ does not appear free in $f$.
-It boils down to the statement that two functions are the same if and only if they give the same result for all possible arguments.
```
lierdakil: Okay, I didn't find a good place to mention this (besides here, I mean), but it is important to mention this, because the term η-reduction often gets thrown around when discussing Haskell programs.

untyped-lambda-calculus/article.md
```diff
@@ -146,63 +146,73 @@ This brings us to the discussion of evaluation strategies.
...
 Call-by-need behaves exactly like call-by-name, except that reduction steps that would duplicate computations, don't.
-The trick is to share the computation corresponding to an argument everywhere it appears, such that the shared computation is performed at most once.
+It does that by sharing the computation corresponding to an argument everywhere it appears, so any shared computation is performed at most once.
 
-Expressions that are not reducible using the chosen evaluation strategy are said to be in the _normal form_.
+To sum up, here's a table with all the mentioned strategies and their characteristics:
 
-If a given term has a normal form under both strong strategies, it's the same for both.
-Note that applicative order strategy is strictly weaker than normal order strategy.
-Some terms that normalize under normal order don't under applicative order.
```
lierdakil: > Some terms that normalize under normal order don't under applicative order.

This is a pretty important point, actually. It follows from the rest of the article, but I think it's good to spell it out.

At least, it would be good to mention that normal order/call-by-name are respectively strongly/weakly normalizing, whereas other strategies aren't necessarily.

untyped-lambda-calculus/article.md
```diff
@@ -146,63 +146,73 @@ This brings us to the discussion of evaluation strategies.
...
-A practically significant variation on call-by-name is _call-by-need_, which is, at least in theory, used in Haskell.
+A more practical variation of call-by-name is _call-by-need_, which is, at least in theory, used in Haskell.
 Call-by-need behaves exactly like call-by-name, except that reduction steps that would duplicate computations, don't.
-The trick is to share the computation corresponding to an argument everywhere it appears, such that the shared computation is performed at most once.
+It does that by sharing the computation corresponding to an argument everywhere it appears, so any shared computation is performed at most once.
 
-Expressions that are not reducible using the chosen evaluation strategy are said to be in the _normal form_.
+To sum up, here's a table with all the mentioned strategies and their characteristics:
 
-If a given term has a normal form under both strong strategies, it's the same for both.
```
lierdakil: > If a given term has a normal form under both strong strategies, it's the same for both.

This is also an important point: the result of the computation doesn't change whether we choose to reduce innermost-first or outermost-first; unless the computation diverges under either, that is.

untyped-lambda-calculus/article.md
```diff
@@ -146,63 +146,73 @@ This brings us to the discussion of evaluation strategies.
...
-Note that applicative order strategy is strictly weaker than normal order strategy.
-Some terms that normalize under normal order don't under applicative order.
-Likewise for the weak strategies.
+| Name                 | Outermost- or innermost- first? | Strong or weak? |
+|----------------------|---------------------------------|-----------------|
+| Normal order         | Outermost                       | Strong          |
+| Applicative order    | Innermost                       | Strong          |
+| Call-by-name (eager) | Outermost                       | Weak            |
+| Call-by-value (lazy) | Innermost                       | Weak            |
```
lierdakil: It's the reverse. Call-by-name is lazy and call-by-value is eager.

lierdakil (commented):
Okay, I have, uh, let's call it "objections", to the last edit. Let's discuss.

untyped-lambda-calculus/article.md
NaeosPsy: Oh, yes, sorry. 🤦‍♂️

This is a good example of what can happen when you reverse the order of things you mention: some people can't read at all. 😂

untyped-lambda-calculus/article.md
NaeosPsy: The problem is that since it is mentioned, it will also look odd if it doesn't show up. 😅 

Perhaps we can write Call-by-need (Haskell) 🤔?

untyped-lambda-calculus/article.md
NaeosPsy: I've updated the construction to be parallel and for the explanations of lazy and eager to be next to call-by-name and call-by-value. Hope it is clearer that way. And of course, I've fixed the thing you commented on.

untyped-lambda-calculus/article.md
NaeosPsy: Updated

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,549 @@
...
+| Name                   | Outermost- or innermost- redex first? | Strong or weak? |
+|------------------------|---------------------------------------|-----------------|
+| Normal order           | Outermost                             | Strong          |
+| Applicative order      | Innermost                             | Strong          |
+| Call-by-name (lazy)    | Outermost                             | Weak            |
+| Call-by-value (eager)  | Innermost                             | Weak            |
+
+### Eta-reduction
+
+Speaking of reductions, let's also briefly mention _eta_-reduction ($\eta$-reduction).
```
NaeosPsy, line 224: @lierdakil I put this in a specific subsection. It makes more sense now. You might want to change the text a little, I obviously don't know what I am talking about or why it is relevant to Haskell users. 😂

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,549 @@
...
+
+This expression contains two redexes, itself and $(\lambda y. y)\;z.$
+The former is not contained within any other redex, so it is the outermost of the two.
+The sub-expression does not contain any other redex, so is the innermost.
+
+Reducing the leftmost outermost redex first is called the _normal order strategy_, and reducing the leftmost innermost first is called the _applicative order strategy_.
+
+In other words, the applicative order strategy first reduces all of the arguments (from left to right) and then applies the abstraction, while normal order does the inverse.
+
+Expressions that are not reducible using the chosen evaluation strategy are said to be in the _normal form_.
```
NaeosPsy, line 184: @lierdakil Since it mentions only strong strategies, I decided to put the comparison here. If it is more of a general statement on the superiority of outermost-first reducing, it should be at the end, but then it should also somehow mention them, I think. If it is so, you're welcome to rewrite and put at the end of the strategy section.

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,552 @@
+# A Brief Look at Untyped Lambda Calculus
+
+When talking about Haskell, the term "lambda calculus" often crops up.
+It's a theoretical framework used to define the meaning of computation in many functional languages, such as Haskell, Agda, Idris, etc.
+Understanding lambda calculus can be very helpful when talking about programs written in these languages.
+
+Perhaps surprisingly, understanding lambda calculus can also be helpful in understanding the C++ template metalanguage since it's a functional language in its own right (although it's not, strictly speaking, based on lambda calculus).
```
DanielRrr, line 7: The reader might want to figure out how lambda is helpful in C++ template metalanguage. Are there any references?

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,552 @@
...
+> His lambda calculus is ignored because it is insufficiently C-like.
+> This criticism occurs in spite of the fact that C has not yet been invented.
+>
+> -- James Iry, [A Brief, Incomplete, and Mostly Wrong History of Programming Languages][brief-incomplete-history]
+
+[brief-incomplete-history]: http://james-iry.blogspot.com/2009/05/brief-incomplete-and-mostly-wrong.html
+
+Lambda calculus, initially envisioned as a formal logic system, was developed by [Alonzo Church](https://en.wikipedia.org/wiki/Alonzo_Church) around the 1930s to explore the foundations of mathematics.
+
+The initial formulation had a logical inconsistency known as the [Kleene–Rosser paradox].
```
DanielRrr: Reference

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,552 @@
...
+We can formally define the pure untyped lambda calculus in terms of _abstractions_ and _applications_.
+
+### Abstractions
+
+An abstraction, or a functional abstraction, is a parametric expression, that is to say, a function.
+Lambda calculus only defines univariate (single-variable) functions, but we can easily extend it to include multivariate functions too, as we'll see later.
+
+We introduce new abstractions using the $\lambda$ symbol.
+An abstraction consists of a _head_ and a _body_, separated by the dot ($.$) symbol.
+The head contains the $\lambda$ symbol and the parameter name.
```
DanielRrr: I would say 'the argument name'

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,552 @@
...
+The body is an arbitrary expression.
+For example, $\lambda x. x$ is an abstraction.
+
+Lambda calculus derives its name from the $\lambda$ symbol used in this notation.
+The symbol was most likely chosen arbitrarily.
+However, there's an apocryphal story that the notation was derived from $\hat x$ morphing first into $\Lambda x$ and later into $\lambda x$.
```
DanielRrr: That's not quite arbitrarily. $\hat x$ has been chosen since Church was following Russell's notation from Principia Mathematica, but the editors had no $\hat$ when they were preparing Church's manuscript for publication. You may google that for a good reference.

untyped-lambda-calculus/article.md
```diff
@@ -0,0 +1,552 @@
...
+For example, $f\; x$ is the application of $f$ to $x$.
+Since all abstractions are univariate by definition, there's no special syntax for multivariate application.
+
+You might notice that using whitespace can lead to ambiguity.
+We'll use parentheses where necessary to avoid it.
+Additionally, we will assume that application is left-associative and binds more tightly than abstractions.
+Hence, $f\;a\;b\;c$ is $((f\;a)\;b)\;c$, i.e. $f$ is applied to $a$, then the result is applied to $b$, then to $c$.
+Since application binds more tightly, we don't need parentheses around the body of an abstraction.
+However, we might need parenthesis around the whole abstraction itself.
+We will also eschew parentheses for nested abstractions, e.g. $\lambda x. \lambda y. x\;y$ is $\lambda x. (\lambda y. (x\;y)).$
```
DanielRrr, line 104: The example with both nested abstractions and applications might be helpful as well.

DanielRrr (commented):
Looks great! Here are a couple of tiny remarks

untyped-lambda-calculus/article.md
lierdakil: The reason I'm being wishy-washy here is that I wasn't able to find a good reference for this when I searched a few years back (better than nth-hand rumours and speculation). If you have one, please share. Otherwise, I'm sticking to this version.

untyped-lambda-calculus/article.md
lierdakil: Well, I can point to (Thielecke. "An introduction to C++ template programming", University of Birmingham, 2016), but that's course notes and are not easy to find.

Beyond that, we have λ-calculus-inspired things in Boost, e.g. [the `lambda` metafunction](https://www.boost.org/doc/libs/1_52_0/libs/mpl/doc/tutorial/the-lambda-metafunction.html). Besides, people have a lot of fun researching the isomorphism between C++ templates and λ-calculus, e.g. (Sinkovics, Porkoláb. "Expressing C++ template metaprograms as lambda expressions." In TFP’09, Jun, pp. 2-4. 2009), (Gozillon, Keir. "Point-Free Templates." In European LLVM Developers Meeting. 2018), and there are a few more, and it is pretty illuminating in its own right (there's a [blog post](https://lettersfromtheoutershell.xyz/2020/10/17/templates-calculus) on this topic too)

Not sure if it makes sense referencing any of those.

untyped-lambda-calculus/article.md
lierdakil: I think that's going to be a bit overkill -- the rules are ultimately pretty simple, and getting too caught up in the examples will break the flow.

untyped-lambda-calculus/article.md
NaeosPsy: Yeah, feels like it would be too much.

untyped-lambda-calculus/article.md
NaeosPsy: Perhaps we can just cut the "the symbol was most likely chosen arbitrarily. However,"?

untyped-lambda-calculus/article.md
lierdakil: Eh. So here's the deal: I've added the reference to https://plato.stanford.edu/entries/paradoxes-contemporary-logic/#IncoCertFormLogi, but it was removed in 5690f94 by @NaeosPsy, I'm guessing because it's a bit too opaque for a casual reader.

I don't really care either way, it's not a journal publication where every statement has to be either referenced or proved, but I'd like to avoid editing wars :sweat_smile:

untyped-lambda-calculus/article.md
lierdakil: Fair enough, changed.

untyped-lambda-calculus/article.md
NaeosPsy: You still have a link to Kleene-Rosser paradox, right? I just didn't believe anyone that read this would care enough for two links. 

You can perhaps link the Plato in the Kleene-Rosser paradox hyperlink if it explains the situation better. 

But yeah. I imagine this section to be something that will be mostly quickly glossed over so there is no need to go in-depth.

untyped-lambda-calculus/article.md
lierdakil: Oh, hey, I did find a better reference. It's a bit inconclusive, though:

> By the way, why did Church choose the notation “λ”?  In [A. Church, 7 July 1964.  Unpublished letter to Harald Dickson] he  stated  clearly  that  it  came  from  the  notation  “x̂”  used  for  class-abstractionby Whitehead and Russell, by first modifying “x̂” to “∧x” to distinguish function-abstraction from class-abstraction, and then changing “∧” to “λ” for ease of printing. This origin was also reported in [J. B. Rosser. Highlights of the history of the lambda calculus.Annalsof the History of Computing, 6:337–349, 1984, p.338].  On the other hand, in his later years Church told two enquirers that the choice was more accidental: a symbol was needed and “λ” just happened to be chosen.

(from History of Lambda-calculus and Combinatory Logic, Felice Cardone, Roger Hindley, 2006)

Also nothing about Church's editor not having a symbol for x̂.

untyped-lambda-calculus/article.md
lierdakil: Okay, let's try this: I've added a journal-style reference in 87137c74d9b2ea5075b6f7d4169037f2b6df4472.

untyped-lambda-calculus/article.md
lierdakil: I've reworded the section slightly to align more with "History of λ-calculus" and added a journal-style reference in 87137c74d9b2ea5075b6f7d4169037f2b6df4472

PR #13, opened 2022-05-28
[PR-1365] Parsing with Haskell (Part 1): Lexing with Alex

heitor-lassarote:
This PR adds an article about using Alex in Haskell. For Happy, see this PR: https://github.com/serokell/blog-posts/pull/14


Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,728 @@
...
+This two-parts tutorial will look into two tools often used together by Haskellers to parse programs, namely Alex and Happy.
+
+Both Alex and Happy are industrial-strength tools powerful enough to parse even Haskell itself.
+At the bottom of this tutorial, you will find links to GHC's lexer and parser if you are curious to see how they look.
+
+This tutorial was written using GHC version 9.0.2, Stack resolver LTS-19.8, Alex version 3.2.7.1, and Happy version 1.20.0.
+
+## Lexing
+
+Before we can start parsing, we should first write a lexer to the grammar. According to A. W. Appel in _Modern Compiler Implementation in ML_ (p. 14):
```
lierdakil: It could be helpful to very briefly introduce the concepts of lexical and grammatical analysis and front-load some alternative terminology (i.e. lexical analyser is also known as "lexer", "scanner" and "tokenizer")

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,728 @@
...
+<!--
+* [X] link to its user guide
+-->
+
+Alex is a Haskell tool to generate lexical analyzers (lexers).
+
+This section aims at giving an introduction to Alex and how you can use it to do useful things in conjunction with Happy (explained later).
+Although Alex and Happy are frequently used together, they are independent tools. They may be combined with other technologies, so you may, for example, use Alex and Megaparsec together if you prefer.
+
+You may also want to read the [Alex User Guide](https://haskell-alex.readthedocs.io/en/latest/).
```
lierdakil: Structurally, I would put this into "further reading". Placing it here implies the reader is expected to leaf through the user guide in order to follow the tutorial, and Alex's user guide is a bit dense.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,728 @@
...
+
+### How it works
+
+<!--
+* [X] make a nifty flowchart
+-->
+
+It will be essential to differentiate between the terms **lexeme** and **token** during this section.
+A lexeme is a valid atom of the grammar, such as a keyword (`in`, `if`, etc.), an operator (`+`, `/`, etc.), an integer literal, a string literal, a left or right parenthesis, an identifier, etc.
+Meanwhile, a token is a class that a specific lexeme belongs to, which will be represented as a Haskell constructor. For instance, `In`, `If`, `Plus`, `Divide`, `Integer 42`, `String "\"foo\""`, `LPar`, `RPar`, and `Identifier "my_function"` are all tokens generated from lexemes.
```
lierdakil: This is not entirely accurate. Token is a pair of token name and (optional) token value. Tokens are generated by the lexer for each lexeme. Token name is the name of a lexical category the lexeme belongs to, while token value is implementation-defined.

You're also contradicting yourself a little here: you say that "token ... will be represented as a Haskell constructor", and meanwhile in the examples you have `Identifier "my_function"` which is certainly not just a constructor.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,728 @@
...
+
+Lexers are often implemented using Deterministic Finite Automata (FDAs), which you can think of as a state machine. For example, supposing that our grammar has the keywords `if` and `in` and also identifiers consisting of only lowercase letters, a small automaton for it could look like so:
```
lierdakil: s/FDAs/DFAs/

> which you can think of as a state machine

DFA _is_ a state machine. More specifically, a deterministic finite-state machine (aka automaton). Moreover, deterministic and non-deterministic finite-state machines (automata) have equal expressive power, so often we don't need to qualify "finite-state machine" at all.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,728 @@
...
+      └───┘     └─┬─┘     └───┘     │
+                  │                 │
+                  │f                │
+                  ▼                 │
+                ┌───┐               │
+                │┌─┐│ [a-z]         │
+                ││3│├───────────────┘
+                │└─┘│
+                └───┘
+```
```
lierdakil, line 195: For someone unfamiliar with automata theory, this is all Greek. Perhaps a good idea to make an aside explaining what this diagram means and how to read it (i.e. that it is a state transition graph for a DFA)

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,728 @@
...
+* 1: Identifier.
+* 2: In.
+* 3: If.
+* 4: Identifier.
+
+Since we stopped at 4, the result is that "int" is an identifier, so we get a token that is `Identifier "int"`.
+The resulting token would be 'In' if we stopped at 2 instead (for the expression "in").
+
+
+What if we consumed no input and stopped at state 0, or the input was something like "Foo123" where there aren't valid transitions?
```
lierdakil: > where there aren't valid transitions

I'll play the role of the insufferable pedant here and complain that DFA by definition can't have a state with no valid transitions. In practice, this means there is a "dead end" state which only has transitions to itself. This "dead end" state is usually omitted on state transition graphs to avoid clutter, but it is always implied.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,728 @@
...
+In such cases, we halt and indicate a lexical error.
+
+We won't write our lexical analyzer from scratch. Instead, we will use Alex to create a list of tokens, which Happy will parse.
```
lierdakil: > will use Alex to create a list of tokens

Technicality, but we will use Alex to generate a DFA-based lexer which will in turn produce the list (or, more generally, a stream) of tokens. Which we'll feed not into Happy itself, but into Happy-generated parser.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,728 @@
...
+
+1. If you run `alex src/Lexer.x`, it will generate `src/Lexer.hs`, where all generated Alex code will be, together with the provided snippets of your code.
+2. Stack automatically generates a file whose path varies between OS and Cabal versions, but on my machine, it's in `.stack-work/dist/x86_64-linux-tinfo6/Cabal-3.4.1.0/build/Lexer.hs`. You can also use `$(stack path --dist-dir)/build/Lexer.hs` to access this file.
+
+For now, at the top of our file, we only declare the module name, but we will soon increase our export list and add more imports.
+
+In the middle section, we first declare our _wrapper_, which indicates the type of code Alex should generate for us (`monadUserState`, which allows us to save custom state) and the input type (`bytestring`, but we could use plain Haskell `String`s instead).
+Consult the Alex User Manual on [Wrappers](https://haskell-alex.readthedocs.io/en/latest/api.html#wrappers) if you want to see other wrappers.
+
+The following section is `tokens :-`, where we will list all the lexemes in our grammar and an optional action on what the lexer should do with the matched lexeme.
```
lierdakil: Okay, so we will _not_ list all the lexemes, that would be in most cases impossible since most interesting languages have a countably infinite lexis. What we _will_ list are _patterns_, defined by regular expressions, which would then match lexemes. 

The "optional action" is not actually optional, as it is what tells Alex what token it should emit for a given pattern. Even "skip" is still an action and is denoted by `;` (which is a synonym for `{ skip }` IIRC). Also, it's not actually optional in the Alex syntax to begin with.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,728 @@
...
+  { start :: AlexPosn
+  , stop :: AlexPosn
+  } deriving (Eq, Show)
+
+data Token = Token
+  { tType :: TokenType
+  , tRange :: Range
+  } deriving (Eq, Show)
+
+data TokenType
```
lierdakil: `TokenType` isn't a great name for this datatype since ostensibly it's not a "token type" (i.e. a named lexical category), but simply a token (i.e. a named lexical category with optional data). My suggestion is to rename this `TokenType` to just `Token` and the `Token` above to `RangedToken`.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,728 @@
...
+The first definition we provided is `<0> $white+ ;`, which simply indicates that all whitespace should be ignored.
+More on this later.
+
+The bottom section contains some boilerplate that Alex requires us to write. These include a data type that needs to be called `AlexUserState`, a value with the initial state called `alexInitUserState`, and a value called `alexEOF`, which instructs Alex how to build the EOF (End-Of-File) token, reached when Alex has finished lexing the input string.
+
+We add some additional datatypes: `Range`, `Token`, and `TokenType`, which we will use throughout the article to describe the tokens that we've successfully created, as well as their positions. Saving the ranges is unnecessary, but it's an excellent addition when reporting errors.
+
+Finally, for the EOF token, we use the `alexGetInput` action to retrieve the current position of the scanner and provide it to the token.
+
+We should provide one token type for each lexeme in the grammar.
```
lierdakil: Again, terminology. We should provide one token _type_ (i.e. name) per lexical category, and one _token_ per lexeme. Lexis in most interesting cases is infinite, and while we can generate individual tokens for all interesting cases at run-time, individual token _names_ per lexeme would pose a problem.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,728 @@
...
+You are free to choose the available tokens. Still, for this tutorial, we will have the following: operators, keywords, literals (strings and integers), identifiers, a token representing the end-of-file (EOF), etc.
```
lierdakil: > You are free to choose the available tokens.

I'm not sure what you wanted to say here, but perhaps you meant "Token names, i.e. lexical categories, for your grammar, are arbitrary, so you're free to define them however you wish"?

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,728 @@
...
+<0> @id     { tokId }
+```
+The syntax `<START_CODE> REGEX { CODE }` means that if Alex has managed to match a token `REGEX`, and it's currently in the `START_CODE` state, it should execute the code `CODE`.
+
+Similarly, for whitespace, we can use `;` instead of giving it an action, which simply means that Alex should do nothing with it.
+
+The idea of start codes is a bit more sophisticated yet valuable.
+Alex works as a state machine, and `0` indicates the initial code in which the machine starts.
+In this case, we can match whitespaces and identifiers while we are in state `0`.
+We will later create other start codes for matching specific things where we wouldn't like to match anything else arbitrarily.
```
lierdakil: Okay, I'm a bit rusty on Alex start codes, but this sounds slightly inaccurate. From what I remember, start codes basically define the initial state of the DFA; The "default" initial state is marked with `0`, but the DFA will then change its state as it works. Hence,

> we can match whitespaces and identifiers while we are in state `0`.

should probably read "we can match whitespace and identifiers when we start from state `0`." (also, "whitespace" doesn't have a plural AFAIK)

> We will later create other start codes for matching specific things where we wouldn't like to match anything else arbitrarily.

This sentence is pretty hard to understand. Perhaps try rephrasing, or just cut the second part of the sentence, i.e. stop at "We will later create other start codes".

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,728 @@
...
+  }
+
+get :: Alex AlexUserState
+get = Alex $ \s -> Right (s, alex_ust s)
+
+set :: AlexUserState -> Alex ()
+set s' = Alex $ \s -> Right (s{alex_ust = s'}, ())
+
+modify :: (AlexUserState -> AlexUserState) -> Alex ()
+modify f = Alex $ \s -> Right (s{alex_ust = f (alex_ust s)}, ())
```
lierdakil: A short aside about alternatively defining `instance MonadState Alex` could be useful.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,728 @@
...
+```
+
+We may also combine multiple regex macros into one, which will be appropriately expanded in its equivalent regular expression.
+For example, here's how we can match an identifier:
+
+```alex
+$digit = [0-9]
+$alpha = [a-zA-Z]
+
+@id = ($alpha | \_) ($alpha | $digit | \_ | ' | \?)*
```
lierdakil: This single quote could use a backslash, if only because common syntax highlighting grammars break on it otherwise:

```
@id = ($alpha | \_) ($alpha | $digit | \_ | \' | \?)*
```

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,728 @@
...
+  , Token {tType = EOF, tRange = Range {start = AlexPn 36 2 14, stop = AlexPn 36 2 14}}
+  ]
+```
+
+#### Lexing strings
+
+Lexing strings should have no surprises. Just stick this in your lexer:
+
+```alex
+<0> \".*\"  { tokString }
```
lierdakil: Oh, no. This is pretty bad, actually. Consider that this matches `"string" id "other string"`. At least use `\"[^\"]*\"`.

lierdakil (commented):
Overall, looks very good, but I have a few nitpicks related to the use of terminology, which is not always precise. I also think that adding a couple asides explaining finite automata could be helpful for a casual reader.

Aside from that, the string regex pattern is too greedy, so it'll produce _unexpected_ results.

Alex and Happy Part 1/article.md
heitor-lassarote: Added it, and explained a bit more about how it works in general.

I reordered this section so "Lexing" appears after the introduction of MiniML as well.

A bit below in the article, I wrote an overview about regexes, which I felt was missing.

Alex and Happy Part 1/article.md
heitor-lassarote: Done.

Alex and Happy Part 1/article.md
heitor-lassarote: Cool, thanks for explaining it. I've made changes to this part, I hope it's accurate now.

Alex and Happy Part 1/article.md
heitor-lassarote: Replaced FDA with DFA and said it's a state machine. I don't understand the second part of your second point very well, do you want me to say that this is just an automaton, without mentioning it's deterministic or finite?

Alex and Happy Part 1/article.md
heitor-lassarote: I made an analogy below, I hope it will be clearer on how one would traverse this diagram.

Alex and Happy Part 1/article.md
heitor-lassarote: Reworded it, and also used "dead end" in the analogy.

Alex and Happy Part 1/article.md
heitor-lassarote: Reworded.

Alex and Happy Part 1/article.md
heitor-lassarote: I've changed it to say we won't list all lexemes and explained about `skip` as well.

Alex and Happy Part 1/article.md
heitor-lassarote: I've renamed it, as well as in the GitHub gists. ~~I didn't update part 2 (about Happy) yet, though, but I will do so soon.~~ Edit: I updated the Happy tutorial and its GitHub gists.

I actually noticed that in the Happy tutorial, I use the version of the lexer without the exercise solutions (i.e., strings are still dumb, and there are no line comments), so I will also update it to include these solutions.

Alex and Happy Part 1/article.md
heitor-lassarote: Changed it.

Alex and Happy Part 1/article.md
heitor-lassarote: I've reworded it, I hope it's clearer now.

Alex and Happy Part 1/article.md
heitor-lassarote: Made the changes, is it better now?

Alex and Happy Part 1/article.md
heitor-lassarote: Added a few words about it.

Now sure why I called it `set` instead of `put`, but I changed it as well.

Alex and Happy Part 1/article.md
heitor-lassarote: Done.

Alex and Happy Part 1/article.md
heitor-lassarote: Oh, that's a good point! I changed it.

Alex and Happy Part 1/article.md
```diff
@@ -58,6 +47,28 @@ let main (unit : ()) : () =
...
+> The lexical analyzer takes a stream of characters and produces a stream of names, keywords, and punctuation marks; it discards white space and comments between the tokens. It would unduly complicate the parser to have to account for possible white space and comments at every possible point; this is the main reason for separating lexical analysis from parsing.
+
+We will use Alex as a tool to generate the lexical analyzer for our grammar.
+It's the first step of the grammatical analysis for our programming language.
+It will take the input stream of characters (a `String`, or in our case, a `ByteString`) representing the program written by the user and generate a stream of tokens (a list), which will be explained more in-depth shortly.
+
+Note that we've mentioned that Alex will _generate_ a lexical analyzer and not that Alex is a lexical analyzer by itself.
+Alex will read a `.x` file created by us specifying how to match lexemes and it will in turn create a `.hs` file which is the generated lexer.
+
+In the second part of this tutorial, the token stream will be consumed by the parser generated Happy; a tool to generate syntactic analyzers.
```
lierdakil: > by the parser generated Happy

Missing "by" before "Happy". Also, while here, this whole sentence doesn't flow very well, but I'll leave tweaking it to experts.

Alex and Happy Part 1/article.md
```diff
@@ -58,6 +47,28 @@ let main (unit : ()) : () =
...
+Happy will try to match tokens according to specific rules that we will describe, where we will produce a tree structure representing a valid MiniML program.
+
+Note that using Alex is not a requirement. You could also use, for example, parser combinators such as Megaparsec for lexing and Happy for parsing if you wanted to.
+
+Making a lexer using parser combinators is pretty doable and manageable.
+Meanwhile, using Alex is a bit more involved, but it has the advantage that performance will be more predictable; besides that, it can be easily integrated with Happy to output one token at a time, which may avoid creating a massive stream of tokens in the memory in the first place.
```
lierdakil: > creating a massive stream of tokens

This is very much a nitpick, but the defining characteristic of a "stream" is it's lazy, so it's not all in memory at once. Perhaps you meant to say "list".

Alex and Happy Part 1/article.md
```diff
@@ -129,9 +138,11 @@ You may also want to read the [Alex User Guide](https://haskell-alex.readthedocs
 
 It will be essential to differentiate between the terms **lexeme** and **token** during this section.
 A lexeme is a valid atom of the grammar, such as a keyword (`in`, `if`, etc.), an operator (`+`, `/`, etc.), an integer literal, a string literal, a left or right parenthesis, an identifier, etc.
-Meanwhile, a token is a class that a specific lexeme belongs to, which will be represented as a Haskell constructor. For instance, `In`, `If`, `Plus`, `Divide`, `Integer 42`, `String "\"foo\""`, `LPar`, `RPar`, and `Identifier "my_function"` are all tokens generated from lexemes.
+Meanwhile, a token consists of the _token name_ and an optional _token value_.
+The token name is the name of a lexical category that the lexeme belongs to, while the token value is implemention-defined.
+We can represent tokens as Haskell constructors with optionally the scanned lexeme. For instance, `In`, `If`, `Plus`, `Divide`, `Integer 42`, `String "\"foo\""`, `LPar`, `RPar`, and `Identifier "my_function"` are all tokens generated for each lexemes.
```
lierdakil: s/as Haskell constructors/as a Haskell sum datatype, where data constructor names correspond to token names and constructor arguments correspond to the token value/

s/each lexemes/their corresponding lexemes/

Alex and Happy Part 1/article.md
```diff
@@ -129,9 +138,11 @@ You may also want to read the [Alex User Guide](https://haskell-alex.readthedocs
...
 
-Lexers are often implemented using Deterministic Finite Automata (FDAs), which you can think of as a state machine. For example, supposing that our grammar has the keywords `if` and `in` and also identifiers consisting of only lowercase letters, a small automaton for it could look like so:
+Lexers are often implemented using Deterministic Finite Automata (DFAs), which is a state machine. For example, supposing that our grammar has the keywords `if` and `in` and also identifiers consisting of only lowercase letters, a small automaton for it could look like so:
```
lierdakil: Uh... grammar. "Automata" is plural, so "state machine" should also be plural I think?

Alex and Happy Part 1/article.md
```diff
@@ -159,27 +170,53 @@ Lexers are often implemented using Deterministic Finite Automata (FDAs), which y
...
+The arrow coming from nowhere to the square with 0 is where you start driving, and squares with another square inside are places where you can park your car. In this case, you are not allowed to park it in the square with 0.
 
-State 0 is the initial state where the automaton begins. Suppose that the automaton can stop at states 1, 2, 3, and 4.
+The arrows also have directions in them, guided by your GPS. Your GPS may tell you to drive straight into an `i`, then turn at an `[a-z]`, and then to stop, for example. However, your GPS may also ask you to take illegal turns, for example, to turn at an `@`, which will always lead you to a dead end, where you will be stuck.
+
+To refine the terminology, the city is the _automaton_ (plural: _automata_) represented by a _state transition graph_, each numbered square is called a _state_, and each arrow is called a _transition_.
+When a transition comes from nowhere and into a state, we say that this state is a _start state_, which is state 0 in this example.
+The places were you can park are called _accept states_, i.e., 1, 2, 3, and 4.
```
lierdakil: A small nitpick, usually I've seen "accepting states", i.e. "states in which the automaton is accepting the input string as lexically correct". I've also seen "final states". I have not yet seen "accept states". Certainly there are sources which use "accept states" too, but judging from Google, it's less common ("final state" being in fact the most common, but I personally find it slightly confusing, since it's not necessarily final)

Alex and Happy Part 1/article.md
```diff
@@ -159,27 +170,53 @@ Lexers are often implemented using Deterministic Finite Automata (FDAs), which y
...
-The resulting token would be 'In' if we stopped at 2 instead (for the expression "in").
+The resulting token would be `In` if we stopped at 2 instead (for the expression "in").
 
-
-What if we consumed no input and stopped at state 0, or the input was something like "Foo123" where there aren't valid transitions?
+What if we consumed no input and stopped at state 0, or the input was something like "Foo123" where we'd reach the implicit state 5?
 In such cases, we halt and indicate a lexical error.
 
-We won't write our lexical analyzer from scratch. Instead, we will use Alex to create a list of tokens, which Happy will parse.
+As started previously, we won't write our lexical analyzer from scratch. Instead, we will use Alex to generate a lexer that creates a stream of tokens, which the Happy-generated parser will parse.
```
lierdakil: s/started/stated/

Alex and Happy Part 1/article.md
```diff
@@ -159,27 +170,53 @@ Lexers are often implemented using Deterministic Finite Automata (FDAs), which y
...
+
+### Regular expressions
+
+Alex uses regular expressions (regexes) to match patterns, we'll keep it pretty simple and explain some of them which we'll use.
```
lierdakil: s/to match patterns/to define patterns/.

For context:

Regular expressions are first and foremost a (regular) pattern definition language. To match a regular expression (i.e. a pattern defined by a regular expression) against some input, you use a DFA (or indeed an NFA), constructed from this regular expression. It can be in fact shown that any regular expression has a corresponding finite automaton that recognizes the language defined by this regular expression (by inductive construction, known as McNaughton-Yamada-Thompson algorithm).

Alex and Happy Part 1/article.md
```diff
@@ -159,27 +170,53 @@ Lexers are often implemented using Deterministic Finite Automata (FDAs), which y
...
+
+A syntax like `[0-9]` means that any digit character between `0` and `9` (inclusive) will be matched. Likewise, `[a-zA-Z]` means that any lowercase or uppercase character can be matched.
+Characters may be excluded as well, for example, `[^\"]` reads as "anything except a double quote mark".
+
+A `*` (called _Kleene star_) means that the expression can be matched zero or more times. Similarly, `+` means that it can be matched one or more times. For instance, `[0-9]+` can match `1234`. You may also find `?`, which means that an expression may match either zero or one time (i.e., it's optional).
+
+A dot (`.`) means "match anything but a newline". You can also use common escape codes like `\n` which will match a newline.
+
+An expression between strings, such as `">="`, means that it should only match exactly that string. An escaped character like `\?` can also match exactly that character.
```
lierdakil: It is not an expression, it's a sequence of characters. Also, "should" and "can" feel a bit too indeterminate for my taste.

I suggest rephrasing to:

> A double-quoted string, such as `">="`, will match exactly the string inside quotes.
> An escaped character like `\?` will match exactly that character.

Alex and Happy Part 1/article.md
```diff
@@ -159,27 +170,53 @@ Lexers are often implemented using Deterministic Finite Automata (FDAs), which y
...
+
+You can group regexes with a space or a pipe (`|`) between them. A space means that Alex should match each regex successively, while the pipe means that it can alternate between choices, such as `0 (x [0-9a-fA-F]+ | o [0-7]+)`, which is a regex that can match hexadecimal or octal numbers.
```
lierdakil: This is a somewhat convoluted explanation of concatenation. First of all, you don't need a space, concatenating regular expressions corresponds to concatenating the languages (i.e. sets of strings) they define (Cartesian-product style). Second, whitespace in Alex regexes is simply ignored (unless it's escaped).

Alex and Happy Part 1/article.md
```diff
@@ -260,23 +297,23 @@ For now, at the top of our file, we only declare the module name, but we will so
...
 
 Finally, for the EOF token, we use the `alexGetInput` action to retrieve the current position of the scanner and provide it to the token.
 
-We should provide one token type for each lexeme in the grammar.
-You are free to choose the available tokens. Still, for this tutorial, we will have the following: operators, keywords, literals (strings and integers), identifiers, a token representing the end-of-file (EOF), etc.
+We should provide one token name for each lexical category in the grammar.
+You are free to add new token names as you wish. Still, for this tutorial, we will have the following: operators, keywords, literals (strings and integers), identifiers, a token representing the end-of-file (EOF), etc.
 
-Our token types will be as such:
+Our token names will be as such:
```
lierdakil: s/.*/Our token datatype is thus:/

Alex and Happy Part 1/article.md
lierdakil: I don't expect anything particular wrt the second part of my comment. I just think it's a neat that DFAs have the same exact power as NFAs (this is not true for more complicated automata). I was teaching a course on compiler theory for a few years, which included elements of language theory, so that's old habits speaking, mostly :sweat_smile: 

BTW, "finite" is actually an important qualifier in theory, infinite state machine would be strictly more powerful. The catch is, it's impossible to implement.

Since you get into some specifics, keeping the "deterministic" qualifier makes sense.

lierdakil (commented):
A couple more nitpicks, mostly related to word choice.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,768 @@
...
+  } deriving (Eq, Show)
+
+data Token
+  = EOF
+  deriving (Eq, Show)
+}
+```
+
+We will use this small template to get Stack to compile the file.
+Don't worry; I will explain everything there soon. :)
```
XJIE6: I tried to understand the code first and after that I read this sentence. So, probably, it is better to put this sentence before the code

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,768 @@
...
+
+Running `stack build --fast --file-watch` will build the project, and Stack will automatically compile the Alex file into a Haskell file.
+
+Alex files are divided into three sections.
```
XJIE6: It wasn't clear for me to separate those three sections in code. So, maybe, it is better to add some comments between sections like 'Top', 'Middle', 'Bottom' or even copy-paste those three bullets as a comments to the code

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,768 @@
...
+You can find the complete code for the lexer until now [here](https://gist.github.com/heitor-lassarote/e56e5330368f2cf9265a6eefa8a7a880).
+
+#### Exercises
+
+1. We support block comments with `(* *)`. Change the scanner to accept line comments, such as `# foo`, which end when they find a newline or EOF.
+
+2. Change the string lexer to accept the following escaped characters: `\\`, `\n`, and `\t`.
+Feel free to add any other escape codes that you can think about.
+
+**Hint**: Create a buffer in `AlexUserState` that adds characters as they are seen and two auxiliary `enterString` and `exitString` functions.
```
XJIE6: Is it possible to hide the hint somehow for people who don't want to read hints?

Alex and Happy Part 1/article.md
heitor-lassarote: I tried to reflow the sentence, I hope it's a bit better now.

Alex and Happy Part 1/article.md
heitor-lassarote: Changed it.

Alex and Happy Part 1/article.md
heitor-lassarote: Thanks for the suggestions, I've done the replacements.

Alex and Happy Part 1/article.md
heitor-lassarote: Accepting states sound better than accept states as well, I've changed it accordingly.

Alex and Happy Part 1/article.md
heitor-lassarote: This is what happens when you send it to review without running through Grammarly for a second time. Done.

Alex and Happy Part 1/article.md
heitor-lassarote: Ah, oops, I remember Thompson's algorithm from my university classes (although I didn't know about the other two people in the name). I had a brain fart to write match instead of define. Thanks for correcting me.

Alex and Happy Part 1/article.md
heitor-lassarote: Replaced it, it feels better now, thank you.

Alex and Happy Part 1/article.md
heitor-lassarote: It didn't occur in my mind that spaces (and the lack of them) would be accepted anywhere in Alex, thanks for pointing it out.

Alex and Happy Part 1/article.md
heitor-lassarote: Done.

Alex and Happy Part 1/article.md
heitor-lassarote: > I was teaching a course on compiler theory for a few years

Wow, that is really awesome. Seems like you're the perfect reviewer for these blog posts, then. :) I hope that the part with Happy will go smoothly, I've written a bit about LR parsing, but I guess I will have to expect many corrections in it. ^^'

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,768 @@
...
+      output <- alexMonadScan
+      if rtToken output == EOF
+        then pure [output]
+        else (output :) <$> go
+```
+
+Startup GHCi and let's check whether this works. It should result in something like this (slightly prettified for ease of visualization):
+
+```haskell
+>>> runAlex "my_identifier" alexMonadScan
```
XJIE6, line 515: I've had several problems running this. Probably we need some more explanations on how to run this. Let's discuss this on call

XJIE6 (commented):
The post is actually pretty good. I didn't fount any major problems

Alex and Happy Part 1/article.md
lierdakil: There's `<details>` tag.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,776 @@
...
+
+A dot (`.`) means "match anything but a newline". You can also use common escape codes like `\n` which will match a newline.
+
+A double-quoted string, such as `">="`, will match exactly the string inside quotes.
+An escaped character like `\?` will match exactly that character.
+
+You can group regexes by concatenating them or by inserting a pipe (`|`) between them.
+
+When you concatenate regexes, such as in `[a-z][A-Za-z0-9]*`, it means that Alex should match each set of strings in sequence.
+One example of a language that can be matched by this regex is `myVariable1`.
```
lierdakil: Either s/language/lexeme/, probably. Or you would properly describe the language, i.e. "the set of all strings that start with a lowercase ASCII letter and have zero or more ASCII letters and numbers following it", but then it's not an example, it's simply the language defined by the regex.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,776 @@
...
+Don't worry; I will explain everything there soon. :)
+
+Running `stack build --fast --file-watch` will build the project, and Stack will automatically compile the Alex file into a Haskell file.
+
+Alex files are divided into three sections.
+* At the top of the file, we define the module and its imports, similarly to Haskell.
+* In the middle, we insert our definitions for the lexer, which will generate the lexemes for our grammar.
+* At the bottom, we may insert more Haskell definitions, such as data structures, auxiliary functions, etc.
+
+The top and bottom sections are provided between the { and }, and all Haskell code will be inlined in the generated Stack file.
```
lierdakil: A good idea to put braces into backtics I think.

```
The top and bottom sections are provided between the `{` and `}`, and all Haskell code will be inlined in the generated Stack file.

```

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,776 @@
...
+At the top of the file, make sure to include the appropriate imports:
+
+```haskell
+import Data.ByteString.Lazy.Char8 (ByteString)
+import qualified Data.ByteString.Lazy.Char8 as BS
+```
+
+### Regex and regex macros
+
+Alex uses regular expressions to match lexemes, so we will use them to recognize the various lexemes that should be available in our grammar.
```
lierdakil: It sounds slightly strange that lexemes are "available in our grammar", I suggest s/that should be available in our grammar/in our language/

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,776 @@
...
+<0> $white+ ;
+
+<0> @id     { tokId }
+```
+The syntax `<START_CODE> REGEX { CODE }` means that if Alex has managed to match a token `REGEX`, and it's currently in the `START_CODE` state, it should execute the code `CODE`.
+
+Similarly, for white space, we can use `;` instead of giving it an explicit action, which simply means that Alex should do nothing interesting with it. This is the same as writing `{ skip }`.
+
+The idea of start codes is a bit more sophisticated yet valuable.
+Alex works as a state machine, and `0` indicates the initial code in which the machine starts.
```
lierdakil: s/ initial code / initial state /

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,776 @@
...
+Let's define our first lexeme based on our identifier:
+
+```alex
+tokens :-
+
+<0> $white+ ;
+
+<0> @id     { tokId }
+```
+The syntax `<START_CODE> REGEX { CODE }` means that if Alex has managed to match a token `REGEX`, and it's currently in the `START_CODE` state, it should execute the code `CODE`.
```
lierdakil: This is slightly misleading. I'd like to suggest s/ and it's currently in the `START_CODE` state / starting from the `START_CODE` state /, but I'm concerned it's going to be somewhat confusing until a couple paragraphs down.

Also, while here, s/match a token/match a pattern/.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,776 @@
...
+
+The `AlexPn` values mean the following, respectively: the number of characters before the token, the line number, and the line column.
+
+Pretty cool stuff.
+
+If you are stuck or want to check that we're on the same page, the code for the lexer up to this section may be found [here](https://gist.github.com/heitor-lassarote/c8f12feab735439a50015b7c60d4f21a).
+
+#### Lexing keywords and operators
+
+We can now lex identifiers, but a programming language typically consists of much more than only this.
```
lierdakil: Perhaps s/only/just/.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,776 @@
...
+<comment> \n   ;
+```
+
+The critical thing to notice is that we can start a comment anywhere, as it has start code `0`, but we can only match a closing comment pair if we begin to match a comment in the first place, thanks to the `comment` start code.
+
+Note that checking for the `*)` lexeme at the start code `0` is unnecessary.
+Without it, it would be emitted as a multiplication followed by a closing parenthesis, but the parser would catch it later.
+Like Haskell, MiniML will support passing operators as first-class functions, like `let add = (+)`.
+However, this has the same deficiency as OCaml: `(*)` is recognized as a comment.
+However, `( *)` is accepted by OCaml but emits a message saying, "Warning : this is not the end of a comment.".
```
lierdakil: Two "however"s in sequence. I suggest rephrasing to:

> OCaml in fact accepts `( *)`, but emits a message saying, "Warning : this is not the end of a comment.".

lierdakil (approved):
I made a few more suggestions, but overall looks pretty good to me already.

Alex and Happy Part 1/article.md
heitor-lassarote: Good suggestions, I changed it accordingly.

Alex and Happy Part 1/article.md
heitor-lassarote: Done!

Alex and Happy Part 1/article.md
heitor-lassarote: I put a few more words about Stack. Please let me know if after updating Stack and using `stack ghci` you still have difficulties running these examples.

Just so that we don't forget, during the call, you suggested placing more anchors in the text, which I've done as well.

Alex and Happy Part 1/article.md
heitor-lassarote: I decided to just replace language with lexeme.

Alex and Happy Part 1/article.md
heitor-lassarote: Done! Not sure how much relevant this part is after the changes, but I decided to keep it in there.

Alex and Happy Part 1/article.md
heitor-lassarote: Yeah, that makes sense. I've changed it.

Alex and Happy Part 1/article.md
heitor-lassarote: I don't think it's too confusing, so I made the changes you suggested.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,817 @@
...
+* [X] generic introduction
+-->
+
+This is the first part of the two parts _Parsing with Haskell_ series.
+Looking for the second part instead? You can find it [here](https://serokell.io/blog/parsing-with-happy).
+
+A critical part of the majority of the compilers is grammar.
+The grammar exposes the compiler's front-end as a convenient way to build programs.
+
+This two-parts tutorial will look into two tools often used together by Haskellers to parse programs, namely Alex and Happy.
```
NaeosPsy: I would mention that you'll show how to parse a small programming language from scratch here.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,817 @@
+# Parsing with Haskell (Part 1): Lexing with Alex
+
+<!--
+* [X] generic introduction
+-->
+
+This is the first part of the two parts _Parsing with Haskell_ series.
+Looking for the second part instead? You can find it [here](https://serokell.io/blog/parsing-with-happy).
+
+A critical part of the majority of the compilers is grammar.
```
NaeosPsy: Are these two lines necessary? I don't believe they add to the introduction, but I might not understand their relevance here.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,817 @@
...
+let the_answer : int =
+  let a = 20 in
+  let b = 1 in
+  let c = 2 in
+  a * c + b * c
+
+let main (unit : ()) : () =
+  print ("The answer is: " + the_answer)
+```
+
```
NaeosPsy, line 48: Suggestion: you parse it into some kind of a syntax tree, right? Perhaps there is a way to show what will be the final result to motivate people. (IDK if it makes sense)

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,817 @@
...
+## Lexing
+
+Before we can start parsing, we should first write a lexer to the grammar, which is also known as _lexical analyzer_, _scanner_, and _tokenizer_.
```
NaeosPsy: Is it known as all of those things or either of those things? Might need "or" instead of "and" at the end of sentence.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,817 @@
...
+According to A. W. Appel in _Modern Compiler Implementation in ML_ (p. 14):
+
+> The lexical analyzer takes a stream of characters and produces a stream of names, keywords, and punctuation marks; it discards white space and comments between the tokens. It would unduly complicate the parser to have to account for possible white space and comments at every possible point; this is the main reason for separating lexical analysis from parsing.
+
+We will use Alex as a tool to generate the lexical analyzer for our grammar.
```
NaeosPsy: Two sentences about Alex together via link feel like it would be right in time.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,817 @@
...
+Although Alex and Happy are frequently used together, they are independent tools. They may be combined with other technologies, so you may, for example, use Alex and Megaparsec together if you prefer.
+
+### How it works
+
+<!--
+* [X] make a nifty flowchart
+-->
+
+It will be essential to differentiate between the terms **lexeme** and **token** during this section.
+A lexeme is a valid atom of the grammar, such as a keyword (`in`, `if`, etc.), an operator (`+`, `/`, etc.), an integer literal, a string literal, a left or right parenthesis, an identifier, etc.
```
NaeosPsy: A lexeme is something you find in the un-lexed code, while a token is something you find in lexed code -- you "lex" lexemes into tokens. 

Would this distinction be true and meaningful? I had to struggle a bit to grasp the difference from this paragraph.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,817 @@
...
+Meanwhile, a token consists of the _token name_ and an optional _token value_.
+The token name is the name of a lexical category that the lexeme belongs to, while the token value is implemention-defined.
+We can represent tokens as a Haskell sum datatype, where data constructor names correspond to token names and constructor arguments correspond to the token value with optionally the scanned lexeme. For instance, `In`, `If`, `Plus`, `Divide`, `Integer 42`, `String "\"foo\""`, `LPar`, `RPar`, and `Identifier "my_function"` are all tokens generated for their corresponding lexemes.
+
```
NaeosPsy, line 164: Seems like you would want some motivation sprinkled on top of here, something goal oriented. 

E.g.

"If your goal is to move from lexemes to tokens, how do you discern different lexemes? For that we use a DFA.  Here's what it is."

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,817 @@
...
+The resulting token would be `In` if we stopped at 2 instead (for the expression "in").
+
+What if we consumed no input and stopped at state 0, or the input was something like "Foo123" where we'd reach the implicit state 5?
+In such cases, we halt and indicate a lexical error.
+
+As stated previously, we won't write our lexical analyzer from scratch. Instead, we will use Alex to generate a lexer that creates a stream of tokens, which the Happy-generated parser will parse.
+
+### Regular expressions
+
+Alex uses regular expressions (regexes) to define patterns, we'll keep it pretty simple and explain some of them which we'll use.
```
NaeosPsy: Is this regular regex syntax? I'm not sure whether it is necessary to explain it beyond a link.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,817 @@
...
+2. Stack automatically generates a file whose path varies between OS and Cabal versions, but on my machine, it's in `.stack-work/dist/x86_64-linux-tinfo6/Cabal-3.4.1.0/build/Lexer.hs`. You can also use `$(stack path --dist-dir)/build/Lexer.hs` to access this file.
+
+For now, at the top of our file, we only declare the module name, but we will soon increase our export list and add more imports.
+
+In the middle section, we first declare our _wrapper_, which indicates the type of code Alex should generate for us (`monadUserState`, which allows us to save custom state) and the input type (`bytestring`, but we could use plain Haskell `String`s instead).
+Consult the Alex User Manual on [Wrappers](https://haskell-alex.readthedocs.io/en/latest/api.html#wrappers) if you want to see other wrappers.
+
+The following section is `tokens :-`, where we will list all the patterns defined by regular expressions to match lexemes in our grammar plus an action on what the lexer should do with the matched lexeme.
+The first definition we provided is `<0> $white+ ;`, which simply indicates that all white space should be skipped.
+More on this later.
```
NaeosPsy: I would cut this line.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,817 @@
...
+### Regex and regex macros
+
+Alex uses regular expressions to match lexemes, so we will use them to recognize the various lexemes in our language.
+Let's start with a simple one: identifiers.
+We are free to choose the specification for identifiers, but let's use the following:
+
+> An identifier is a sequence of alphanumeric characters, primes ('), question marks (?), and underscores (_).
+> The identifier must begin with a letter or an underscore.
+
+Alex allows the creation of _character set macros_, which are shortcuts that you can use to avoid duplicating macros. We use `$NAME = CHARACTER_SET` to define a character set macro.
```
NaeosPsy: Should it perhaps be "duplicating regexes" or "duplicating character sets"?

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,817 @@
...
+
+### Regex and regex macros
+
+Alex uses regular expressions to match lexemes, so we will use them to recognize the various lexemes in our language.
+Let's start with a simple one: identifiers.
+We are free to choose the specification for identifiers, but let's use the following:
+
+> An identifier is a sequence of alphanumeric characters, primes ('), question marks (?), and underscores (_).
+> The identifier must begin with a letter or an underscore.
+
```
NaeosPsy, line 405: Would add something here so that is is more goal-oriented / flows better. 

E.g.

"To make it easier to write the regex pattern for identifiers, we can use character set and regex macros. "

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,817 @@
...
+```
+
+At the top of the file, make sure to include the appropriate imports:
+
+```haskell
+import Data.ByteString.Lazy.Char8 (ByteString)
+import qualified Data.ByteString.Lazy.Char8 as BS
+```
+
+### Regex and regex macros
```
NaeosPsy: There seems to be a bit of a chaotic moving here from regex to identifier to regex to identifier. 

What I would suggest is to call this section Lexing identifiers, and put the text about regex macros in a block like this.
```
<hr>
**Regex macros**

text
<hr>
```
Here's an example of how that looks: https://serokell.io/blog/untyped-lambda-calculus#defining-computation

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,817 @@
...
+It will be useful to know how `AlexInput` looks like as well:
+
+```haskell
+type AlexInput = (AlexPosn,    -- current position,
+                  Char,        -- previous char
+                  ByteString,  -- current input string
+                  Int64)       -- bytes consumed so far
+```
+
+Let's create our `tokId` function. We do so by inserting this definition at the bottom of the file.
```
NaeosPsy: Is it bottom **part** of the file, perhaps?

Alex and Happy Part 1/article.md
NaeosPsy: Gist checkpoints could be helpful, perhaps?

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,817 @@
...
+  ]
+```
+
+We would like to detect such cases and also support nested comments.
+We will create two helper functions, `nestComment` and `unnestComment`, which will keep track of how many layers of comments we have and allow us to detect whether we reached the end of the file with an unclosed comment.
+
+In addition, Alex provides the `andBegin` combinator that will execute an expression and change the current start code.
+
+```alex
+<0> $white+ ;
```
NaeosPsy, line 687: I thought this was nice way to show off start codes. :)

Alex and Happy Part 1/article.md
lierdakil: Yes and no. Alex uses mostly standard notation, but it has a few quirks.

Alex and Happy Part 1/article.md
lierdakil: > A lexeme is something you find in the un-lexed code, while a token is something you find in lexed code -- you "lex" lexemes into tokens.

Eh, the last part ("lex" lexemes into tokens) is _kind of_ correct, hence lexer is also called "tokenizer", but this grinds my gears a little.

Lexeme is a fancy word for "word", i.e. a lexical unit, an element of the [lexis](https://en.wikipedia.org/wiki/Lexis_(linguistics)) (i.e. vocabulary), something you can meaningfully categorize within your grammar; natural language examples are: "cat" is a lexeme of class "noun", "an" is a lexeme of class "indefinite article", etc. In computer science we use it kind of vaguely since we often include punctuation into the lexis, which linguists don't usually do.

A token is a representation of a lexeme used internally by lexer and parser, generally agreed to consist of a _token name_, a tag identifying the corresponding lexeme's category, and _token value_, which is arbitrary data (implementation defined, but often it's just a string that contains the lexeme itself, or some other representation thereof).

These two exist in somewhat parallel realities, and certainly lexemes don't "disappear" after lexing, we just represent them as tokens in the code.

Alex and Happy Part 1/article.md
heitor-lassarote: I'd prefer to keep this section. I don't know how it's usually done, but in my university, I remember we learned regex during our compiler classes for the lexing part, so I think it makes sense to explain it to the user here.

Alex and Happy Part 1/article.md
heitor-lassarote: I added a few more words about it, I hope it makes it a little clearer.

Alex and Happy Part 1/article.md
heitor-lassarote: Added a link to its website and said a bit more, hope it's better.

Alex and Happy Part 1/article.md
heitor-lassarote: I've added the resulting AST at the end, I'm not sure though if it's a very motivating example. Maybe in the future, we should write a tutorial about writing a tree-walking interpreter perhaps continuing from this tutorial, perhaps?

Alex and Happy Part 1/article.md
heitor-lassarote: Yeah, I wanted to write something better but I'm bad at introductions. :pepe_sad: I removed it.

Alex and Happy Part 1/article.md
heitor-lassarote: I will add a few more checkpoints later.

Alex and Happy Part 1/article.md
NaeosPsy: Sure, that's ok.

Alex and Happy Part 1/article.md
NaeosPsy: Well, haha, ASTs are scary but I think it's better than nothing. 

Perhaps we can add that this can be easily (😂) compiled or interpreted. 


Regarding article: I wouldn't mind that. :)

Alex and Happy Part 1/article.md
```diff
@@ -365,16 +381,20 @@ import Data.ByteString.Lazy.Char8 (ByteString)
...
 We are free to choose the specification for identifiers, but let's use the following:
 
 > An identifier is a sequence of alphanumeric characters, primes ('), question marks (?), and underscores (_).
 > The identifier must begin with a letter or an underscore.
 
-Alex allows the creation of _character set macros_, which are shortcuts that you can use to avoid duplicating macros. We use `$NAME = CHARACTER_SET` to define a character set macro.
+<hr>
+<b>Regex macros</b>
+
+To make it easier to write the regex pattern for identifiers, we can use character set and regex macros.
```
NaeosPsy: Sorry for bothering, but I think this sentence needs to be outside so people understand for what the box is there. 😁

NaeosPsy:
I changed the structure a little bit to get rid of some duplicate explanations of Alex (of which I was the cause but.. 😂). Now all of the subsections of the  "tutorial part" are wrapped in one H2 -- Lexing with Alex. I think it makes more sense like this.

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,842 @@
...
+  (ELetIn _ (Dec _ (Name _ "a") [] Nothing (EInt _ 20))
+  (ELetIn _ (Dec _ (Name _ "b") [] Nothing (EInt _ 1))
+  (ELetIn _ (Dec _ (Name _ "c") [] Nothing (EInt _ 2))
+  (EBinOp _ (EBinOp _ (EVar _ (Name _ "a")) (Times _) (EVar _ (Name _ "c"))) (Plus _) (EBinOp _ (EVar _ (Name _ "b")) (Times _) (EVar _ (Name _ "c")))))))
+, Dec _ (Name _ "main") [Argument _ (Name _ "unit") (Just (TUnit _))] (Just (TUnit _))
+  (EApp _ (EVar _ (Name _ "print")) (EPar _ (EBinOp _ (EString _ "\"The answer is: \"") (Plus _) (EVar _ (Name _ "the_answer")))))
+]
+```
+
+n.b.: The `_` will contain the range of the parsed tree, we've omited it here for brevitity.
```
NaeosPsy: I have a feeling like it should be the ranges of the parsed tokens, since every token has a range, am I wrong?

Alex and Happy Part 1/article.md
```diff
@@ -0,0 +1,842 @@
...
+  [ RangedToken {rtToken = LPar, rtRange = Range {start = AlexPn 0 1 1, stop = AlexPn 1 1 2}}
+  , RangedToken {rtToken = Times, rtRange = Range {start = AlexPn 1 1 2, stop = AlexPn 2 1 3}}
+  , RangedToken {rtToken = Identifier "my", rtRange = Range {start = AlexPn 3 1 4, stop = AlexPn 5 1 6}}
+  , RangedToken {rtToken = Identifier "comment", rtRange = Range {start = AlexPn 6 1 7, stop = AlexPn 13 1 14}}
+  , RangedToken {rtToken = EOF, rtRange = Range {start = AlexPn 13 1 14, stop = AlexPn 13 1 14}}
+  ]
+```
+
+We would like to detect such cases and also support nested comments.
+We will create two helper functions, `nestComment` and `unnestComment`, which will keep track of how many layers of comments we have and allow us to detect whether we have reached the end of the file with an unclosed comment.
```
NaeosPsy, line 681: People might go searching for these functions, would be good to give them a little hint that you're building up to them.

Alex and Happy Part 1/article.md
heitor-lassarote: You are correct that every token will have a range, but the output here is a tree, and these ranges will correspond to tree nodes.

NaeosPsy (approved):

PR #15, opened 2022-06-06
[PR-1299] What's That Typeclass: Functor

alyoanton9:


whats-that-typeclass-functor/article.md
```diff
@@ -0,0 +1,467 @@
+You might already know the `map` function for lists:
+
+```haskell
+> map (+1) [1, 2, 3]
```
sras: If your audiance does not know what a `Functor` is, chances are that they are not really familiar with stuff like `(+1)`, might be even be not familiar with lambdas. I think it is better stick to separate function with explicit type signatures, but here, using a lambda might be a good enough compromise.

whats-that-typeclass-functor/article.md
```diff
@@ -0,0 +1,467 @@
...
+[2, 3, 4]
+```
+
+It takes a function of type `a -> b` and applies it to each element of a list of type `[a]`. The values change, but the context (`[]`) remains the same.
```
sras: > The values change, but the context (`[]`) remains the same.

This seems a bit confusing. May be we can start by saying some types can have other types inside of them, and a `map` operation acts on the inner type, leaving the outer type the same, or something like that. This notion of a 'context' might be a bit jarring for a beginner. I am not sure, but It also seems a bit ambigious, because IIRC I have seen this "context" used in a lot of monad tutorials to refer the monads environment. So I think using it here is better avoided.

whats-that-typeclass-functor/article.md
```diff
@@ -0,0 +1,467 @@
...
+
+Transforming a value inside a fixed context like that is common in Haskell.
```
sras: > is common in Haskell.

But isn't this operation common in all languages? And Haskell and the likes just provides ways to generalize it?

whats-that-typeclass-functor/article.md
```diff
@@ -0,0 +1,467 @@
...
+<details>
+<summary>The type of <code>map'</code> modifying <code>Maybe</code></summary>
+
+```haskell
+map' :: (a -> b) -> Maybe a -> Maybe b
+```
+<hr>
+</details>
+
+Finally, we can write a `map'` for an arbitrary data type `f` with one type argument `a` — `f a`. By analogy, can you figure out how it would look like?
```
sras: It might be better if we admit here upfront that this `map` function cannot work without knowing the innards of the type it is acting upon (and hence the reliance on a Typeclass), Or the reader might get that "magical Haskell" feeling here, making them lose all hope of ever groking the concept of the Functor.

whats-that-typeclass-functor/article.md
```diff
@@ -0,0 +1,467 @@
...
+-- Note that only the second value is being fmap'ed
+> fmap chr (65,65)
+(65,'A')
+> fmap chr ('a',97)
+('a','a')
+```
+
+As you may have guessed, `map` is just a synonym of `fmap` for list. But it can do much more.
+
+With `fmap`, we can reverse strings, convert numbers to strings, perform algebraic and boolean operations, and do other stuff with values inside data types like `[]`, `Maybe`, `Either a`, pair `((,) a)`, and others.
```
sras: >With `fmap`, we can reverse strings, convert numbers to strings, perform algebraic and boolean operations,

It is a bit unclear to me what this means. Did you mean that we can apply arbitrary functions/transformation to wrapped types using `fmap`?

whats-that-typeclass-functor/article.md
```diff
@@ -0,0 +1,467 @@
...
+
+Functor provides methods for structure-preserving transformations.
+
+You’re already familiar with `fmap`. However, just any `fmap` is not enough for a correct Functor instance. Some Haskell typeclasses have laws — conditions that instances have to meet. Only this guarantees that the instance is correct.
+
+Functor, in particular, needs `fmap` to adhere to the following laws.
+
+---
+
+### Laws of Functor
```
sras: When I was new to Haskell, I remember coming across these "laws" and wondering about the following things.

1. What happens if I implement some instance that does not obey these laws?
2. Are these laws somehow enforced by the language?
3. Where is this laws written authoritatively? 
4. Is it possible to implement instances that break these laws?

Would be nice if we can address some of these here...

whats-that-typeclass-functor/article.md
```diff
@@ -0,0 +1,467 @@
...
+</details>
+
+Exactly! `(<$)` just runs `fmap` with `const` function.
+
+
+We got to know the basic components of Functor, and now it’s time to create our own instance of this type class!
+
+## Creating an instance of Functor
+
+Let’s see how you can implement your own instance of Functor. We’ll use the [`NonEmpty`](https://hackage.haskell.org/package/base-4.16.1.0/docs/Data-List-NonEmpty.html#t:NonEmpty) data type as an example.
```
sras, line 267: > We’ll use the [`NonEmpty`]

It would be nice to use a type with a normal constructor so that the example is more accessable.

whats-that-typeclass-functor/article.md
```diff
@@ -0,0 +1,467 @@
...
+  fmap _ []   = []
+```
+
+In fact, the principle is the same — splitting a list on head and tail, and calling the function recursively. On each iteration, we apply `f` to the head and call `fmap` with the tail of the list.
+
+In the end, we need to cover the empty-list case — ‘fmapping’ an empty list gives an empty list.
+
+It’s done! You’ve just seen how to define your own instance of the Functor typeclass. You can practice this more with the exercises section.
+
+## Functor is not a container only
```
sras: I think this is better phrased as "A Functor is not always a container".

whats-that-typeclass-functor/article.md
```diff
@@ -0,0 +1,467 @@
...
+5
+> (+1) . (*2) $ 2
+5
+```
+
+In fact, it’s uncommon to consider a function a Functor in Haskell, it’s more of a fancy case. However, you’ll not be puzzled when you come across magic and strange usage of `fmap` in the future!
+
+## Why do you need to know Functor?
+
+In general, Functor is not a really powerful typeclass. Its methods are easy to grasp and to implement.
```
sras: >Functor is not a really powerful typeclass. Its methods are easy to grasp and to implement.

Not sure about this statement. I don't think the methods being easy to grasp have anything to do with the usefulness of the abstraction. Is the methods in Monad typeclass much more complicated than the Functor methods? Nevertheless, Monads are pretty powerful, Right?

whats-that-typeclass-functor/article.md
NaeosPsy: Could NonEmpty be written as a plain product type instead?

whats-that-typeclass-functor/article.md
NaeosPsy: That's a good way to introduce that that matches with how we talk about fmap, I like that.

whats-that-typeclass-functor/article.md
NaeosPsy: True. 

I guess the phrase "we can write a map for arbitrary data type" is in this sense, wrong?

NaeosPsy: We can generalize this type signature for an arbitrary data type `f` with one ... 


^ Perhaps this would be better for time being?

whats-that-typeclass-functor/article.md
NaeosPsy: Oh, that's interesting. 

I think the readers should have enough experience (+ this is kind of intuitive), but a lambda is OK as well. 

On a side note, I believe we can expect around 99% of audience to know what a lambda is or be able to recognize it: it's just so common in 2022.

sras: Oh, I have a general preference for keeping everything dead simple, except for the thing that is being explained. I guess I am a bit burned by tutorials that set out to explain something, and use somethings equally obscure as part of the explanation.

For example, back when I wanted to know about data families in Haskell, this is the page I will usually end up at https://wiki.haskell.org/GHC/Type_families, and in that page, at the begining of the explanation, there is something like. 

> As an example, consider Ralf Hinze's [generalised tries](http://www.cs.ox.ac.uk/ralf.hinze/publications/GGTries.ps.gz), a form of generic finite maps.

And I don't have any clue about "tries" let alone "Relf Hinze's generalized" ones, and it seems to me that this aspect makes it only accessable to people coming from some sort of academic background.

So I think if you are explaining something that needs explaining, it is a good idea to keep everything else dead simple, and I think it will considerable expand the size of the audiance that your explanation will work for.

whats-that-typeclass-functor/article.md
sras: >I guess the phrase "we can write a map for arbitrary data type" is in this sense, wrong?

I think it could be made more accurate. We cannot write a `map` function for any type, but for just those that has defined how to `fmap` over themselves (via `Functor` Typeclass). So `Functor` typeclass is the generalization, and `map` function can work for many types because it uses this generalized interface.

whats-that-typeclass-functor/article.md
sras: Yea, I guess so. Something like

```
data NonEmpty a = NonEmpty { neHead :: a, neTail :: [a] }
```
Right?

whats-that-typeclass-functor/article.md
NaeosPsy: Yes, you are entirely correct on that. I try to operate Serokell Blog articles on exactly the same assumption.

whats-that-typeclass-functor/article.md
NaeosPsy: It does mention that it is for lists on top.

whats-that-typeclass-functor/article.md
cognivore: > And Haskell and the likes just provides ways to generalize it?

I wouldn't say that it generalises anything, to be fair. Languages with wonkier semantics have more expressive transformers of data! (Because looser semantics means more expressivity, at a cost of correctness).

I would think of Haskell as a language that provides taxonomies for such transformers of data!

whats-that-typeclass-functor/article.md
serokellcao: I would also provide a `:t map`, specialised to `[a]`.

whats-that-typeclass-functor/article.md
serokellcao: Haskell doesn't quite generalise them. Languages with wonkier semantics have more general transformers.

Haskell, however, provides taxonomies for such transformers.

whats-that-typeclass-functor/article.md
```diff
@@ -0,0 +1,467 @@
...
+- understand how to define and use your own Functor instances;
+- see why functors are useful and where they could be used;
+- be able to consolidate your knowledge with exercises.
+
+## How to generalize `map`
+
+Look at the type signature of `map`:
+
+```haskell
+map :: (a -> b) -> [a] -> [b]
```
serokellcao, line 50: Should be higher, I think. Exactly in this form. Here, perhaps, we recap. You mention part of the signature eralier without spelling it out.

whats-that-typeclass-functor/article.md
serokellcao: To be pedantic: we can also write `map` for type class `Extract` which has a method `extract :: w a -> a`. Sure, `Comonad` is a `Functor`, but just `extract` can be defined without something being a covariant Functor. For example 

```
newtype ToInt a = ToInt {unToInt :: (a -> Int)}
```

has `extract = unToInt`, but no `Functor`. I think that these are the intricacies we should convey to the reader. I value this discussion a lot.

whats-that-typeclass-functor/article.md
```diff
@@ -0,0 +1,467 @@
...
+(65,'A')
+> fmap chr ('a',97)
+('a','a')
+```
+
+As you may have guessed, `map` is just a synonym of `fmap` for list. But it can do much more.
+
+With `fmap`, we can reverse strings, convert numbers to strings, perform algebraic and boolean operations, and do other stuff with values inside data types like `[]`, `Maybe`, `Either a`, pair `((,) a)`, and others.
+
+The implementation of `fmap` is data-type-specific. So to use it on a data type, that data type needs to have an implementation for it.
```
serokellcao: This seems to be at odds with a statement "you can implement map for an arbitrary type".

whats-that-typeclass-functor/article.md
```diff
@@ -0,0 +1,467 @@
...
+
+Functor, in particular, needs `fmap` to adhere to the following laws.
+
+---
+
+### Laws of Functor
+
+1. **Identity**
+
+    `fmap id == id` — applying `id` function to the wrapped value changes nothing.
```
serokellcao: @sras suggested to remove currying. Perhaps we remove currying here too?

whats-that-typeclass-functor/article.md
serokellcao: I would add an important observation that for some classes there are multiple Functors. For example, for list we can just fmap at most over its head, while still complying with the laws. Same for Monoid, btw.

whats-that-typeclass-functor/article.md
```diff
@@ -0,0 +1,467 @@
...
+> 1 <$ [0]
+[1]
+> 1 <$ [1..5]
+[1,1,1,1,1]
+
+-- replace each element
+-- of the list `[Int]`
+-- with a `Char`
+-- Note: the resulted type
+-- is `String` as `type String = [Char]`
```
serokellcao: I would rather not promote `String = [Char]` because it's a misfeature. 
I would rather have something like `[] <$ [1..5]`.

whats-that-typeclass-functor/article.md
```diff
@@ -0,0 +1,467 @@
...
+
+But is there any data that corresponds to the function? The answer is yes, it’s `(->)`:
+
+```haskell
+> :info (->)
+type (->) :: * -> * -> *
+data (->) a b
+```
+
+`(->)` is parametrized with two types — `(->) a b`. In other words, it is `a -> b` — a function with one argument.
```
serokellcao: This is a very good section! 👏

whats-that-typeclass-functor/article.md
```diff
@@ -0,0 +1,467 @@
...
+
+```haskell
+-- fmap :: (Int -> Int) -> (Int -> Int) -> Int -> Int
+> fmap (+1) (*2) 2
+5
+> (+1) . (*2) $ 2
+5
+```
+
+In fact, it’s uncommon to consider a function a Functor in Haskell, it’s more of a fancy case. However, you’ll not be puzzled when you come across magic and strange usage of `fmap` in the future!
```
serokellcao: We absolutely must give Contravariants a treatment here, perhaps even before showing that (->) is a Functor.

serokellcao: Contravariants are famously can accept isomophisms in lens (see contramapping :: Contravariant f => AnIso s t a b -> Iso (f a) (f b) (f s) (f t)).

I also think that we need to mention Bifunctor too here while talking about Tuples, *before* we talk about (->).

I finally think that we need to make a follow-up post about Profunctor ~> Strong, Choice as we build to "the Arrow tutorial" post :D

serokellcao (changes requested):

whats-that-typeclass-functor/article.md
NaeosPsy: Bruh, this is a beginner article. 😁

whats-that-typeclass-functor/article.md
sras: >we can also write map for type class Extract which has a method

Yea, I meant there should be some code that provides information or an implements an interface that can be used to do the mapping.

whats-that-typeclass-functor/article.md
sras: >fmap at most over its head

But how can you leave the head and tail with potentially different types?

whats-that-typeclass-functor/article.md
alyoanton9: I've edited this sentence a bit and added a note about context

I also believe there are a lot of articles about monads calling them context. However, it's not some formal notation or documented thing, but the most suitable word explaining what monads do. The same for functor, it's just a word that helps to grasp the idea. So replacing it with another one less appropriate, that doesn't resonate with monads explanation, looks to me like just comlicating the things

whats-that-typeclass-functor/article.md
alyoanton9: Changed a bit. Does it look clearer and more fair now?

whats-that-typeclass-functor/article.md
alyoanton9: Rephrased to avoid inaccurate wording. In addition, the note like following could be added here:

> Finally, we can generalize type signature of the `map'` for some* data type `f` with one type argument `a` — `f a`.

...

\* Not every data type could be a functor ...

However, I'm not sure if it's worth doing 🤔

whats-that-typeclass-functor/article.md
alyoanton9: Yeah, it is

I just wanted to add the concrete examples, resonating with the ones above

whats-that-typeclass-functor/article.md
alyoanton9: Rephrased a bit

whats-that-typeclass-functor/article.md
alyoanton9: I've added it to the first code snippet

whats-that-typeclass-functor/article.md
alyoanton9: I've added the related comment [here](https://github.com/serokell/blog-posts/pull/15#discussion_r906841807)

whats-that-typeclass-functor/article.md
NaeosPsy: Love the rewording!

whats-that-typeclass-functor/article.md
NaeosPsy: @alyoanton9 remember to change the docs as well :)

whats-that-typeclass-functor/article.md
NaeosPsy: I think the problem is that we're showing an unimplementable map'. 

The real type sig is this, right? 

fmap :: Functor f => (a -> b) -> f a -> f b

I think once we introduce this, we can make everything clear. 🤔

whats-that-typeclass-functor/article.md
NaeosPsy: As I wrote in another comment, we need to introduce the the type of fmap and somehow explain what happens there:

fmap :: Functor f => (a -> b) -> f a -> f b

And that would solve these discussions.

NaeosPsy: It would be nice to mention that you cannot really implement a polymorphic fmap, you need a typeclass to guard and provide type-specific implementations, hence we have Functor. 🤔

whats-that-typeclass-functor/article.md
NaeosPsy: I guess this could be more readable, I'll try to suggest a rephrase.

whats-that-typeclass-functor/article.md
NaeosPsy: It would be good to address these above the explanation of functor Laws, especially because for most people this will be the first mathy typeclass.

whats-that-typeclass-functor/article.md
NaeosPsy: I like the phrase "provides taxonomies" tbh. 

I will leave this open for now since the whole intro is a bit shaky, needs rewording/re-arranging of sentences, which I'll look at.

whats-that-typeclass-functor/article.md
alyoanton9: I am not sure if I got it correctly

@serokellcao, do you propose to change this example in the following way?
```haskell
-- replace each element
-- of the list `[Int]`
-- with a `Char`
> '1' <$ []
[]
> '1' <$ [0]
['1']
> '1' <$ [1..5]
['1', '1', '1', '1', '1']
```
Or do you mean that it's better to remove this example completely, because any mentioning `type String = [Char]`  is undesired? 🤔 

Anyway, it seems strange to me to hide this misfeature from readers. The very first usage of GHCi would promote `String = [Char]`, so it looks unfair to avoid this in the article.

whats-that-typeclass-functor/article.md
alyoanton9: Rewrote

whats-that-typeclass-functor/article.md
sras: I think this section "Why do you need to know Functor", can be reduced to something that amounts to "Because it enables one to apply transformation on the wrapped type without knowing anything about the wrapper, and it is a very useful thing".

whats-that-typeclass-functor/article.md
NaeosPsy: Yeah, it would be better not to touch charlists. 

Given that the example would work with 1 of the number type, I would prefer that.

whats-that-typeclass-functor/article.md
alyoanton9: I added a [small section](https://github.com/serokell/blog-posts/commit/f47a71dbca41039eed9cdc3acfa33cb233318e82#diff-4d5bf4bf274bf0b54f99958d24e2d64f6cefc8743f5c0c21150e83148c2e000dR140) about typeclass laws in Haskell

> I would add an important observation that for some classes there are multiple Functors. For example, for list we can just fmap at most over its head, while still complying with the laws. Same for Monoid, btw.

@serokellcao do you mean that we can just define orphan instances in different modules? 🤔 Or is there something else?..

whats-that-typeclass-functor/article.md
alyoanton9: [Added](https://github.com/serokell/blog-posts/commit/f47a71dbca41039eed9cdc3acfa33cb233318e82#diff-4d5bf4bf274bf0b54f99958d24e2d64f6cefc8743f5c0c21150e83148c2e000dR369)

whats-that-typeclass-functor/article.md
alyoanton9: Done

whats-that-typeclass-functor/article.md
NaeosPsy: I've rephrased in my edit commits.

whats-that-typeclass-functor/article.md
NaeosPsy: The intro has been reworded, so I'll mark this as resolved now.

NaeosPsy:
@alyoanton9 @sras 

I've made some edits. Some of those involved adding new sentences and creating new sections + deleting some of the written stuff to make it less dense for beginners. I invite you to read what we have and say what you think. 🙈

I believe we're quite close to shipping.

whats-that-typeclass-functor/article.md
NaeosPsy: Since we added a section on typeclasses, and the multiple functors is not that relevant to the main concept (as it is with Monoid, where we cover this), I will resolve this.

whats-that-typeclass-functor/article.md
```diff
@@ -0,0 +1,507 @@
...
+
+It’s uncommon to consider a function a Functor in Haskell. In fact, it’s more of a fancy case. However, you’ll not be puzzled when you come across this magical and strange usage of `fmap` in the future!
+
+## Why is Functor useful?
+
+In general, Functor is not an extraordinary typeclass. Its methods are easy to grasp and to implement.
+
+Nevertheless, a Haskell project can hardly do without a couple of Functor instances. It enables one to do transformations on the wrapped type without knowing anything about the wrapper, and that's very beneficial.
+
+Moreover, Functor is a solid basis and the predecessor of the Applicative typeclass, which further leads to monads. Hence, it brings you one step closer to the dream of many – understanding monads.
```
sras: > it brings you one step closer to the dream of many – understanding monads.

I think we should not feed into this idea that monad is such a mystical and hard thing. We can write something like "which further leads to monads, which is a widely used Typeclass in Haskell and understanding it gives you considerably greater command of the language"

PR #18, opened 2022-07-12
Added Wasp on list of haskell open source projects

Martinsos:
I was just reading your blog post about open-source Haskell projects and thought Wasp would fit well into it (I am one of the creators of Wasp)!


NaeosPsy:
Hi! 

Wasp looks cool. 👍

If you don't mind, I'll put a mention about Wasp in the compilers section of the article instead of it having a separate entry.

Martinsos:
Thanks @NaeosPsy , sounds good!

NaeosPsy:
Great, I've added it to the article then. 👍

PR #23, opened 2022-07-19
[PR-1427] Structs in Rust

sancho20021:


structs-in-rust/article.md
```diff
@@ -0,0 +1,366 @@
...
+    int y;
+}
+```
+
+### Unit struct
+Unit struct is a trivial example that contains zero fields:
+```
+struct A;
+```
+Don't be afraid if you see such strange struct definitions. You may wonder what is the purpose of unit structs when they don't contain any information. In reality, they are being used (see [std::fmt::Error](https://doc.rust-lang.org/std/fmt/struct.Error.html), [chrono::offset::Utc](https://docs.rs/chrono/0.4.19/chrono/offset/struct.Utc.html)), as they can have methods and different trait implementations. Both things I just mentioned will be discussed further in the article.
```
NaeosPsy: This isn't very convincing, could you come up with a simple example?

structs-in-rust/article.md
```diff
@@ -0,0 +1,366 @@
...
+//           ^^^
+//           Expected type, found keyword `mut`.
+    immutable: bool,
+}
+```
+There are two options:
+- Have an immutable binding to the struct. Then you cannot change any of its fields.
+- Have a mutable binding, in which case you can set whatever field you choose.
+
+### Convenient field init syntax
```
NaeosPsy: Wait, how does this differ from regular initialization?

structs-in-rust/article.md
```diff
@@ -0,0 +1,366 @@
...
+//  ^^ struct update syntax
+};
+```
+`..b1` tells Rust that we want to initialize the rest of `b2`'s fields with the values from `b1`.
+
+#### Field Init Shorthand
+It is possible to omit *field* in `field: value` initialization syntax if the value is a variable (or function argument) with a name that matches the field:
+```
+fn new_bicycle(brand: String) -> Bicycle {
+    let kind = String::from("road");
```
NaeosPsy: I'd move this to the argument, it's simpler that way.

structs-in-rust/article.md
```diff
@@ -0,0 +1,419 @@
...
+let point: Point = Default::default();
+let double_bool: DoubleBool = Default::default();
+
+assert_eq!(point, Point { x: 0, y: 0 });
+assert_eq!(double_bool, DoubleBool(false, false));
+```
+
+The most important lines in this snippet are those where we create `point` and `double_bool`. Because both `Point` and `DoubleBool` implement `Default`, we can invoke the `default()` method of the `Default` trait in each case to generate an empty value of type `Point` and `DoubleBool` respectively.
+
+*Exercise: try to remove explicit type annotations from lines 3 and 4 and see what the compiler tells you.*
```
NaeosPsy: I'd add the word "Why?" at the end to make it feel more like an exercise.

structs-in-rust/article.md
```diff
@@ -0,0 +1,419 @@
...
+
```
NaeosPsy, line 332: It seems that you could show how to manually implement a trait here as well, seems like there is space?

structs-in-rust/article.md
sancho20021: Ok. I think the example would require generics and of course traits. What do you think of placing the example after the `Traits` section?

structs-in-rust/article.md
sancho20021: Ah, I see, misleading header. I made a change in the latest commit which should fix the problem

structs-in-rust/article.md
sancho20021: Ok

structs-in-rust/article.md
sancho20021: Ok

structs-in-rust/article.md
sancho20021: @NaeosPsy But not defining our own traits? Just implementing exising?

NaeosPsy: Yeah.

structs-in-rust/article.md
NaeosPsy: That feels too complicated for this. 🤔 Can it be textual -- a description of what would you do.

sancho20021: I want to show smth like this:
```
mod library {
    pub trait ErrorLogger {
        fn error(error_code: i32);
    }

    pub fn function_with_errors<L: ErrorLogger>() {
        L::error(25);
    }
}

mod my_app {
    use super::library::*;

    // Logger 1
    struct PoliteLogger;

    impl ErrorLogger for PoliteLogger {
        fn error(error_code: i32) {
            println!(
                "Sorry for interrupting you, but an error with code {} occurred.",
                error_code
            );
        }
    }

    // Logger 2
    struct SimpleLogger;

    impl ErrorLogger for SimpleLogger {
        fn error(error_code: i32) {
            println!("Error: {}", error_code);
        }
    }

    fn option1() {
        function_with_errors::<PoliteLogger>();
    }

    fn option2() {
        function_with_errors::<SimpleLogger>();
    }
}
```

So do I need to describe the idea in words? But we assume the reader doesn't know that functions can take generic arguments.

sancho20021: What about dropping unit structs? I can't come up with a simple explanation without leaving the reader with questions

NaeosPsy: Yeah, that is possible as well. 👍

sancho20021: I decided to remove it)

structs-in-rust/article.md
sancho20021: I added a small example

structs-in-rust/article.md
```diff
@@ -0,0 +1,443 @@
...
+- Associated functions. They don't require an instance to be called on. Like static functions in Java.
+- Methods. They are invoked on an instance of a type.
+
+### Self
+
+You may have noticed that methods have `self` as their first argument. It represents the instance of the struct the method is being called on, similar to how it is done in Python. Some syntax sugar is involved here. Let's desugar it in two steps.
+
+|     |                                              |                                             |
+| --- | -------------------------------------------- | ------------------------------------------- |
```
Vlad-Shcherbina, line 370: (optional) For completeness, maybe also desugar
```
// a method that consumes struct instance
fn foo(self)
```

Or did you omit it intentionally?

structs-in-rust/article.md
```diff
@@ -0,0 +1,443 @@
...
+#### Get
+
+You can get the desired value using the well-known dot notation.
+
+```
+let x = p_named.x;
+let y = p_named.y;
+```
+
+One may wonder: what is the syntax to address a specific field of a tuple struct? Fortunately, Rust has chosen the simplest way possible, by enumerating tuple components from *0* to *n*.
```
Vlad-Shcherbina: Don't know what you mean by *n*, but maybe should be *0* to *n - 1*.

structs-in-rust/article.md
```diff
@@ -0,0 +1,443 @@
...
+```
+struct Struct {
+    mutable: mut i32,
+//           ^^^
+//           Expected type, found keyword `mut`.
+    immutable: bool,
+}
+```
+
+There are two options:
```
Vlad-Shcherbina: (this comment is optional; totally up to you if it's worth mentioning)

There is a third option, kind of: make the field private and don't expose methods that modify it.
It will be still possible to change that field by assigning to the whole struct.

structs-in-rust/article.md
```diff
@@ -0,0 +1,443 @@
...
+
+A trait is similar to an *interface* in Java or a *typeclass* in Haskell. It defines certain functionality that we might expect from some type.
+
+Usually, traits declare a list of methods that can be called on the types that implement this trait.
+
+Here is a list of commonly-used traits from the standard library:
+
+- [`Debug`](https://doc.rust-lang.org/std/fmt/trait.Debug.html). Used to format (and print) a value using `{:?}`.
+- [`Clone`](https://doc.rust-lang.org/std/clone/trait.Clone.html). Used to get a duplicate of a value.
+- [`Default`](https://doc.rust-lang.org/std/default/trait.Default.html). Used to create a default instance of type.
```
Vlad-Shcherbina: I'm not sure about `Default`. All other traits here are very important. `Default` is really not.

Is it worth bringing it up?

structs-in-rust/article.md
```diff
@@ -0,0 +1,443 @@
...
+}
+
+let p = Point::new(0, 1);
+
+// format a pair
+assert_eq!(format!("{:?}", p), String::from("Point { x: 0, y: 0 }"));
+assert_eq!(format!("{}", p), String::from("(0, 0)"));
+```
+
+The syntax seems pretty natural. We `impl`ement the trait `std::fmt::Display` `for` the struct called `Point`. In the opened code block we then have to define every function that is declared in the trait. In our case, there are only one — `fmt`.
```
Vlad-Shcherbina: I think there is a general style advice to avoid phrases like "it's easy to see, that...".
If it's easy to see, the reader will notice that it's easy to see themselves.
If it's not, it's a lie.

Maybe drop "the syntax seems pretty natural" and proceed directly to the explanation.

structs-in-rust/article.md
```diff
@@ -0,0 +1,443 @@
...
+
+One may wonder: what is the syntax to address a specific field of a tuple struct? Fortunately, Rust has chosen the simplest way possible, by enumerating tuple components from *0* to *n*.
+
+```
+let x = p_unnamed.0;
+let y = p_unnamed.1;
+```
+#### Set
+
+It is possible to change the value of a field using the same dot notation, but the struct instance **must be defined as mutable**.
```
Vlad-Shcherbina: Maybe "struct variable" instead of "struct instance"?
"Struct instance" sounds like a value in memory, and mutability is a property of bindings and references, not values.

structs-in-rust/article.md
```diff
@@ -0,0 +1,443 @@
...
+impl std::fmt::Display for Point {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "({}, {})", self.x, self.y)
+    }
+}
+
+let p = Point::new(0, 1);
+
+// format a pair
+assert_eq!(format!("{:?}", p), String::from("Point { x: 0, y: 0 }"));
```
Vlad-Shcherbina: You can compare `String` with `&str`, so it could be simply
```
assert_eq!(format!("{:?}", p), "Point { x: 0, y: 0 }");
```

structs-in-rust/article.md
```diff
@@ -0,0 +1,443 @@
...
+Just like in Java or C++, in Rust, you can define **methods** for a particular type. This is done in an `impl` block.
+
+```
+impl Point {
+    // Associated function
+    fn new(x: i32, y: i32) -> Point {
+        Point { x, y }
+    }
+
+    // Immutable method
```
Vlad-Shcherbina: I don't think terms "immutable method" and "mutable method" are common. Sounds kind of inaccurate, I dunno.
I wish there were succinct phrases for that, but nothing comes to mind.

If we are trying to avoid the term "borrow" for simplicity, the clunky alternative is "a method that can't modify struct instance" and "a method that can modify struct instance".

structs-in-rust/article.md
```diff
@@ -0,0 +1,443 @@
...
+`Self` is an alias for the type of the impl block. In our case, it is `Point`.
+
+Methods can mutate the struct instance they are associated with, but this requires `&mut self` as the first argument. This way they can be called only via mutable binding.
+
+### Why use methods instead of regular functions?
+
+There are at least two reasons.
+
+1. Nice-looking dot notation.
+2. Code organization. It is a lot more convenient for the user of your library to see all functionality of the struct inside one impl block instead of tens of functions spread all across the module.
```
Vlad-Shcherbina: I agree it's about code organization, but disagree with the explanation. Nothing prevents one from keeping regular functions together textually, or from scattering methods across multiple impl blocks all over:
```
struct MyStruct {}

...

impl MyStruct {
  fn do_stuff
}

...

impl MyStruct {
  fn do_other_stuff()
}
```

--------

There two things:
* You don't have to import methods:
    ```
    use my_module::MyStruct;

    ...
    let s = MyStruct::new();
    s.do_stuff();
    s.do_other_stuff();
    ```
    vs
    ```
    use my_module::{MyStruct, do_stuff, do_other_stuff};  // ugly

    ...
    let s = MyStruct::new();
    do_stuff(&s);
    do_other_stuff(&s);
    ```
* Methods (and associated functions in general) are namespaced. So you don't have name collisions and can write
    ```
    my_rectangle.area()
    my_circle.area()
    ```
    instead of more verbose and redundant
    ```
    rectangle_area(my_rectangle)
    circle_area(my_circle)
    ```

structs-in-rust/article.md
```diff
@@ -0,0 +1,443 @@
...
+| 3   | `fn has_same_x(self: &Point, other: &Point)` | `fn shift_right(self: &mut Point, dx: i32)` |
+
+
+`Self` is an alias for the type of the impl block. In our case, it is `Point`.
+
+Methods can mutate the struct instance they are associated with, but this requires `&mut self` as the first argument. This way they can be called only via mutable binding.
+
+### Why use methods instead of regular functions?
+
+There are at least two reasons.
```
Vlad-Shcherbina: Another tiny advantage of methods is autoreferencing:
You can write
```
point.shift_right(3);
```
instead of explicit
```
(&mut point).shift_right(3);
```

For functions, you always have to reference:
```
shift_point_right(&mut point, 3);
```

structs-in-rust/article.md
```diff
@@ -0,0 +1,443 @@
...
+
+1. Nice-looking dot notation.
+2. Code organization. It is a lot more convenient for the user of your library to see all functionality of the struct inside one impl block instead of tens of functions spread all across the module.
+
+## Methods vs traits
```
Vlad-Shcherbina: I dunno, is it worth going into this in a beginner-level post?

A beginner is unlikely to create their own traits, so the rule is simple:
* if you are implementing a common functionality for which a trait exists (converting to string, comparison, ...), maybe `impl` that trait
* if you are doing something specific to your application, probably use methods in regular `impl` blocks

Vlad-Shcherbina (commented):
Overall this seems correct, writing style is precise and simple.

All my comments are optional, you are the author.

structs-in-rust/article.md
NaeosPsy: I agree. 👍

structs-in-rust/article.md
NaeosPsy: I think the rule is good and can be added.

The rest of it should probably stay, as much for beginners as for SEO.

structs-in-rust/article.md
sancho20021: I think I omitted it because:
- I didn't show examples with `self`
- Don't want to explain `self` vs `&self`

structs-in-rust/article.md
sancho20021: Rephrased

structs-in-rust/article.md
sancho20021: Okay, I'll leave it as it is, but thanks for the idea, didn't think about it!

structs-in-rust/article.md
sancho20021: Removed

structs-in-rust/article.md
sancho20021: Yes, thanks for the advice!

structs-in-rust/article.md
sancho20021: Done

structs-in-rust/article.md
sancho20021: Ok

structs-in-rust/article.md
sancho20021: Rephrased as suggested

structs-in-rust/article.md
sancho20021: Done

structs-in-rust/article.md
sancho20021: Added

structs-in-rust/article.md
sancho20021: Added

NaeosPsy:
@sancho20021 I've added copy edits. Please check if they don't break anything. They do break the "methods" section -- I invite you to fix it by fixing the TODO's that I've left.

sancho20021:
@NaeosPsy I've fixed the TODO's. Fixes are in the first commit.

Please, make sure you review the part about `Self`, because I rewritten the sentence.

And I also suggest a small reordering in the last commit, moving the chapter about `Self` up, right after the `&mut self` part, so the reader gets the explanation immediately. What do you think?

NaeosPsy:
Ok @sancho20021 , looks great, thank you very much! 🙏

I'll schedule it to be published next Tuesday, 16th August.

sancho20021:
Great, thank you too!

PR #25, opened 2022-08-12
[PR-1422] Typed lambda calculus

lierdakil:
There are things I had to drop for volume and time, so the second half of the article might feel a little bit like a bus tour. Honestly, not sure if it's better to leave it as is or get into the weeds on each topic. Fair warning: getting into the weeds could easily inflate the article size at least 3 times, and it's not like it's small to begin with.

I would really like for someone more proficient in type theory than me to give this a look -- I did get in a bit farther than my usual depth in places, and I can't say I'm standing very firmly there, i.e. I'm not certain there aren't stupid glaring mistakes lurking (I hope not, but yeah, a bit out of my depth to be entirely sure)


typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,996 @@
...
+Finally, in 1908 he came up with the notion of type theory.
+Therein, he introduced a hierarchy of types, and requiring that "no totality can contain members defined in terms of itself", essentially forbidding this paradoxical definition outright.
+
+This seemed very promising at the time, so Russel together with Alfred Whitehead went to work on axiomatizing the foundations of mathematics.
+What resulted from it is the infamous Principia Mathematica, a three-volume tome of arcane mathematics.
+To get a very general idea what that looks like, the proof that $1+1$ is, in fact, $2$, is only completed by page 86 in the second volume.
+More volumes were in fact planned, but after 1913 the enthusiasm seemingly ran out.
+
+It's arguably a good thing it did though, as in 1931 Kurt Gödel came out with his incompleteness theorem, proving that arithmetic can not prove its own consistency.
+This essentially has put an end to the Hilbert's program.
```
NaeosPsy: We talked about this here as well, as I understand: https://serokell.io/blog/formal-verification-history 

Is there a way to tie a link to this here somewhat?

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,996 @@
...
+***
+
+</details>
+
+### Typing relations, rules and systems
+
+To express such restrictions, we introduce the so-called _typing relation_, written as $t : T$, read as "term $t$ has type $T$".
+We also introduce _types_, which can be thought of as sets of terms of certain general form.
+
+<p><details>
```
NaeosPsy: I think this can just be a sentence in text, no need for the details block here.

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,996 @@
...
+<summary>On types as sets intuition</summary>
+
+Turns out (Reynolds, 1984), this naive interpretation doesn't hold to rigorous scrutiny.
+Still, it's a good enough mental model unless you're doing type theory research.
+
+***
+</details></p>
+
+Using typing relations, we thus assign a type to each term.
+Rules which we use to assign types are called _typing rules_ and together types and typing rules constitute a _type system_.
```
NaeosPsy:
```suggestion
Rules which we use to assign types are called _typing rules_.
Together, types and typing rules constitute a _type system_.
```

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,996 @@
...
+\begin{array}{c}x:\sigma\in\Gamma\\\hline \Gamma\vdash x:\sigma\end{array} & \text{variable} \\
+\begin{array}{c}\Gamma\vdash t_0:\tau\to\tau';\quad\Gamma\vdash t_1:\tau\\\hline \Gamma\vdash t_0\;t_1:\tau'\end{array} & \text{application} \\
+\begin{array}{c}\Gamma,\;x:\tau\vdash t:\tau'\\\hline \Gamma\vdash \lambda x.\; t:\tau\to\tau'\end{array} & \text{abstraction} \\
+\begin{array}{c}\Gamma\vdash t_0:\sigma;\quad\Gamma,\;x:\sigma\vdash t_1:\tau\\\hline \Gamma\vdash \mathbf{let}\;x = t_0\;\mathbf{in}\; t_1 : \tau\end{array} & \text{let-in} \\
+\begin{array}{c}\Gamma\vdash t:\sigma';\quad\sigma'\sqsubseteq\sigma\\\hline \Gamma\vdash t : \sigma\end{array} & \text{instantiation} \\
+\begin{array}{c}\Gamma\vdash t:\sigma;\quad\alpha\notin free(\Gamma)\\\hline \Gamma\vdash t : \forall \alpha.\;\sigma\end{array} & \text{generalization} \\
+\end{array}
+$$
+
+This is just one possible formulation of the ruleset, other equivalent formulations exist, but this one is enough for our purposes for now.
```
NaeosPsy:
```suggestion
This is just one possible formulation of the ruleset.
Other equivalent formulations exist, but this one is enough for our purposes for now.
```

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,996 @@
...
+Can it be introduced into typed lambda calculi in a more or less principled fashion?
+The answer, as it turns out, is yes.
+This is shown by Philip Wadler and Stephen Blott in (Wadler, 1989a).
+
+The idea here is to make polymorphism _constrained_.
+That is, instead of simply a $\forall \alpha$ quantifier, we may have a $\forall \alpha. P\;\alpha,$ where $P$ is some predicate.
+
+With type classes specifically, predicate $P$ can essentially only be "does this type belong to an open class of types".
+
+A type class defines an "interface", a set of functions that can be used with any type in the class, but the class itself usually provides no implementation.
```
NaeosPsy: Is it possible to put a link to our Haskell typeclass article somewhere here? (https://serokell.io/blog/haskell-typeclasses) Or would it not be good?

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,996 @@
...
+In the paragraph above, I'm glossing over a lot of details for simplicity.
+In particular, I'm conflating typing and translation to HM, which are two distinct steps.
+Still, this gets the core idea accross well enough.
+
+***
+
+</details></p>
+
+To work around this issue, it is required that $\mathbf{over}$ and $\mathbf{inst}$ must have global scope.
+That's why in Haskell we can't have local instance declarations.
```
NaeosPsy: Given that there is an association between GHC and Haskell (so it feels a bit confusing), and the details feel wordy here, I'd propose something like this perhaps: 

> in X, we can have local instance declarations only through a lot of dark and arcane compiler magic.

NaeosPsy (commented):
Read the text from start to finish to get acquainted, just a couple of comments that came up.

NaeosPsy (commented):
Read the text from start to finish to get acquainted, left a couple of comments that came up.

I like this the way it is, seems like a pretty great bus tour. If you want to expand on something, probably a standalone article for that thing would be better.

typed-lambda-calculus/article.md
lierdakil: Uh... Haskell is a standard, GHC is one implementation (with a lot of extensions).

But really, there's no way to get local instances even with GHC extensions, getting those needs some `unsafeCoerce` finagling using knowledge of GHC's runtime representations. So it's "kind of" possible with a lot of magic™.

lierdakil:
Apologies, I apparently failed to commit the first editing pass before creating the PR, so grammar (and even syntax) was wonky in places, not to mention questionable phrasing, and I've added some clarifications in a couple places. Committed as a fixup just now.

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,1016 @@
...
+
+Typed lambda calculus is simply untyped lambda calculus with types.
+The exact features of the type system used can be chosen mostly freely, and we will discuss some common choices in this article.
+
+Since we've discussed untyped lambda calculus in the previous article, now we'll focus more on the applications of type theory.
+We'll also assume knowledge of basic lambda calculus syntax and concepts introduced in the previous article.
+
+With the untyped lambda calculus, we have discussed unrestricted computation.
+We also noted that the untyped lambda calculus is hard to use as a logic system due to the paradoxes it creates.
+Adding Russel's type theory to lambda calculus makes it, under certain conditions, logically consistent, which in turn makes it possible to mathematically guarantee certain behaviors (or lack thereof).
```
int-index: Russell has two L's in his name, here and in other occurrences in the article.

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,1016 @@
...
+
+Since the early 2000s, type theory research has mostly focused on the homotopy type theory (HoTT).
+In HoTT the notion of equivalence is generalized to homotopy -- i.e. the topological notion of similarity.
+HoTT turns out to be powerful enough to discuss the foundations of mathematics again and is used in the univalent foundations project, which is arguably somewhat of a spiritual successor to Hilbert's program.
+
+## Simply typed lambda calculus
+
+Untyped lambda calculus contains some "puns". In the previous article, we noted that the term $fls$ (denoting boolean zero in Church encoding) and the term $c_0$ (denoting natural zero in Church encoding) have the same representation.
+
+Furthermore, untyped lambda calculus allows for constructions like, for example, $plus\; tru\; c_2,$ which don't have a defined semantic meaning.
```
int-index: 1. What does the semicolon mean?
2. "semantic meaning" is "wet water"

Not really sure what you're trying to say here, so I cannot suggest a correction.

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,1016 @@
...
+
+Various typed lambda calculi introduce some restrictions on the form that terms in different expressions might have, as to avoid this punning and meaningless terms.
+
+### Lambda calculus with atomic boolean values
+
+To facilitate the discussion, let us introduce a couple additional atomic value-terms to our lambda calculus syntax, namely, values $true$ and $false$.
+This, of course, makes our lambda calculus impure.
```
int-index: Given that "impure" usually means "having unrestricted mutation and IO", this sentence could easily confuse the reader.

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,1016 @@
...
+
+The reason we need to do this will become apparent later, but for now, suffice it to say that we can't have simply typed pure lambda calculus as the "simple" type system is not expressive enough.
```
int-index: I'm not a fan of the "you'll see later" style of explanation. My attention span is one paragraph. Just say what you have to say or don't say anything at all.

Not to mention that your previous paragraph starts with "to facilitate the discussion", so you already told me to suspend disbelief and roll with whatever you're about to introduce.

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,1016 @@
...
+
+Since we introduced atomic values, we also need to introduce an eliminator (i.e. consumer) for these values, in this case let's consider an $\mathbf{if} \ldots \mathbf{then} \ldots \mathbf{else} \ldots$ operator.
+
+The semantics of this operator should be reasonably obvious.
```
int-index: This article is an introduction. Even the word "semantics" itself is not obvious! If you were to introduce the semantics of `if ... then ... else ...` properly, you'd extend the reduction relation with two elimination rules and a congruence rule. Do you actually believe that the reader is capable of inferring that if this is their first exposure to the topic?

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,1016 @@
...
+We'll only note that there are two meaningful options here: either both branches are computed (corresponding to "eager" evaluation) or only the matching one is (corresponding to "lazy" evaluation).
+In practice, programming languages generally prefer the second option for obvious reasons, but the exact choice doesn't affect the following discussion.
+
+So, we'd like to avoid meaningless terms.
```
int-index: I have no idea whatsoever of what you mean when you say "meaningless terms".

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,1016 @@
...
+What terms would be meaningful in this context?
+For one, we expect the first argument to be either $true$ or $false$.
+For two, we'd like both branches to have the same general "form" to avoid complications with nesting this construct into other computations.
+
+Formally speaking, "meaningless" terms are called "stuck" -- the term is not a value with defined semantic meaning, and further reduction is impossible.
```
int-index: Now you say "stuck", but you do not explain what "stuck" means. Really, this is something you should've started with to motivate introducing a type system in the first place. We want to avoid stuck terms.

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,1016 @@
...
+To facilitate the discussion, let us introduce a couple additional atomic value-terms to our lambda calculus syntax, namely, values $true$ and $false$.
+This, of course, makes our lambda calculus impure.
+
+The reason we need to do this will become apparent later, but for now, suffice it to say that we can't have simply typed pure lambda calculus as the "simple" type system is not expressive enough.
+
+Since we introduced atomic values, we also need to introduce an eliminator (i.e. consumer) for these values, in this case let's consider an $\mathbf{if} \ldots \mathbf{then} \ldots \mathbf{else} \ldots$ operator.
+
+The semantics of this operator should be reasonably obvious.
+We'll only note that there are two meaningful options here: either both branches are computed (corresponding to "eager" evaluation) or only the matching one is (corresponding to "lazy" evaluation).
+In practice, programming languages generally prefer the second option for obvious reasons, but the exact choice doesn't affect the following discussion.
```
int-index: Not obvious to me. I mean, yeah, to avoid unnecessary computation — but then why stop at boolean operators? Why not make all computations short-circuit? In fact, Haskell does exactly that. There must be a reason why the majority of programming languages opts for lazy evaluation of boolean operators but nothing else. It's probably very subtle and not obvious at all.

And if it doesn't affect the discussion, just don't mention it.

typed-lambda-calculus/article.md
lierdakil: `\;` is a space in LaTeX.  Without it, we'd have _plustruc₂_; with it, _plus tru c₂_

The point is, addition of boolean true and natural 2 (both Church-encoded) has no defined semantics (both are introduced in the previous article). One could of course define something, but that's not the point. The point is, untyped λ-calculus allows for such "meaningless" terms, and we want to avoid that for hopefully obvious reasons.

More formally, in an impure λ-calculus rather than "meaningless" terms we would have stuck terms, i.e. terms that are neither values nor could be reduced further. But in this motivating example, introducing impure calculus would be more confusing than helpful.

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,1016 @@
...
+What terms would be meaningful in this context?
+For one, we expect the first argument to be either $true$ or $false$.
+For two, we'd like both branches to have the same general "form" to avoid complications with nesting this construct into other computations.
+
+Formally speaking, "meaningless" terms are called "stuck" -- the term is not a value with defined semantic meaning, and further reduction is impossible.
+
+<details>
+<summary>On the types of conditional branches</summary>
+
+Turns out, that if we want branches to have potentially different types, we need types to depend on terms.
```
int-index: I am waiting for you to define what a type is, and you're already talking about dependent types. Only to immediately dismiss them.

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,1016 @@
...
+
+## Simply typed lambda calculus
+
+Untyped lambda calculus contains some "puns". In the previous article, we noted that the term $fls$ (denoting boolean zero in Church encoding) and the term $c_0$ (denoting natural zero in Church encoding) have the same representation.
+
+Furthermore, untyped lambda calculus allows for constructions like, for example, $plus\; tru\; c_2,$ which don't have a defined semantic meaning.
+
+Various typed lambda calculi introduce some restrictions on the form that terms in different expressions might have, as to avoid this punning and meaningless terms.
+
+### Lambda calculus with atomic boolean values
```
int-index: After having read this section, I have no idea what I was supposed to take away from it. I am not a step closer to understanding typed lambda calculi, which is what the article purports to explain.

typed-lambda-calculus/article.md
lierdakil: The terminology, again, was introduced in the previous article. λ-calculus can't really have side effects to begin with.

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,1016 @@
...
+Since the early 2000s, type theory research has mostly focused on the homotopy type theory (HoTT).
+In HoTT the notion of equivalence is generalized to homotopy -- i.e. the topological notion of similarity.
+HoTT turns out to be powerful enough to discuss the foundations of mathematics again and is used in the univalent foundations project, which is arguably somewhat of a spiritual successor to Hilbert's program.
+
+## Simply typed lambda calculus
+
+Untyped lambda calculus contains some "puns". In the previous article, we noted that the term $fls$ (denoting boolean zero in Church encoding) and the term $c_0$ (denoting natural zero in Church encoding) have the same representation.
+
+Furthermore, untyped lambda calculus allows for constructions like, for example, $plus\; tru\; c_2,$ which don't have a defined semantic meaning.
+
```
NaeosPsy, line 36: Just a quick suggestion: would the analogy of binary be meaningful / help here? You also can do operations on random binary values, but types are what give meaning to them. Most people reading, will, most likely, be more familiar with this.

typed-lambda-calculus/article.md
NaeosPsy: I agree with Vlad. While we did cover what is "pure" in the previous article, and we should, more or less, expect them to have read it, I'd expect a non-significant portion of readers to be people whose exposure to lambda calculus is something like the first chapter of Haskell from First Principles. So I'd throw them an explanation, link, or think if the statement is even necessary here.

typed-lambda-calculus/article.md
int-index: For some reason the GitHub renderer displays the semicolon as is, but ok, hopefully our blog does better in this regard.

> i.e. terms that are neither values nor could be reduced further

How do you decide whether a stuck term is a value or not? I don't see you defining the set of values anywhere, so there are no "meaningless" terms. There are stuck terms and there are redexes, and you don't get to talk about meaning at all.

typed-lambda-calculus/article.md
int-index: Of course it can have side effects. You just introduce primitive operations that have some "effect on the outside world" that is not a part of your formal semantics.

And it doesn't matter if you introduced confusing terminology in the previous article, it is still confusing.

lierdakil:
As a heads-up, I'm not taking Vlad's criticism particularly well today. It's not without merit. But I'm going to abstain from discussing this at least until tomorrow. Cheers.

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,712 @@
+# A Look at Typed Lambda Calculus
+
+This is the second article in our short series on lambda calculus.
+You can find the first article [here](https://serokell.io/blog/untyped-lambda-calculus).
+
+We've also [previously covered](https://serokell.io/blog/formal-verification-history) some history of the formal verification and thus type theory.
+Reading that article might provide historical background for what we'll be discussing here.
+
+Typed lambda calculus is simply untyped lambda calculus with types.
```
int-index: Something that might not be obvious to the reader (and this introduction is a good place to clear this up) is that there isn't a single typed lambda calculus. We are actually talking about various typed lambda calculi: simply-typed lambda calculus, System F, the calculus of constructions, etc.

And even the STLC isn't a single system, we can freely choose the ground types. 

If you don't want to spell this out explicitly, using an indefinite article could do the job
```suggestion
A typed lambda calculus extends untyped lambda calculus with a type system.
```

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,712 @@
...
+The exact features of the type system used can be chosen mostly freely, and we will discuss some common choices in this article.
+
+Since we've discussed untyped lambda calculus in the previous article, now we'll focus more on the applications of type theory.
```
int-index: It's up to you, but this paragraph could be made shorter and more actionable:

"If you are unfamiliar with lambda calculus, the [first article](https://serokell.io/blog/untyped-lambda-calculus) is a better starting point. Now we focus on adding types to it."

Another option is to merge it with the paragraph that already contains the link. Most people don't read long introductions anyway, so it's better to have just a few paragraphs with useful links and get to the topic ASAP.

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,712 @@
...
+We'll also assume knowledge of basic lambda calculus syntax and concepts introduced in the previous article.
+
+With the untyped lambda calculus, we have discussed unrestricted computation.
```
int-index: I'm a bit slow on the uptake, but so are many other readers. I have not realized straight away that this was supposed to be a recap of the previous article, and I simply got swamped with unfamiliar terms: "unrestricted computation", "Russell's type theory", "logically consistent". Easily fixed by merging this with the previous paragraph:

```suggestion
We'll also assume knowledge of basic lambda calculus syntax and concepts introduced in the previous article: unrestricted computation, logic system, paradox, Russell's type theory, logical consistency.
```

Then I immediately know that I'm supposed to read the first article to educate myself, a list of keywords to guide me. Though I also wonder if all of these concepts are truly required to understand STLC.

int-index (commented):
A few comments about the introduction. Need to go now, will try to take a look at the rest of the article later.

typed-lambda-calculus/article.md
lierdakil: I've tried to clean up the introduction, roughly along the lines you suggest here.

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,711 @@
...
+The most common argument for the strongly-typed languages is that if a program type checks, it will never segfault (or have any other type-related issues), but I think this argument sells typed languages short.
+A sufficiently strict and expressive type system allows for encoding much more than simply "not segfaulting".
+In some cases, the whole domain can be expressed in types.
+
+In this article, we're concerned with the basics, for the most part, so we won't get too much into the more expressive type systems like System FC or systems with dependent types.
+
+## Lambda calculus with atomic boolean values
+
+We will now introduce atomic values to the lambda calculus.
+While it is feasible to construct a value representation using only λ-abstractions (using Church encoding or otherwise) for almost any value we might be interested in, practically speaking, it's not very convenient.
```
int-index: Convenience wouldn't be my first concern.

The more pressing problem with these encodings is that the computational complexity of many operations is different from what you'd get with an actual data structure. Not important for booleans, but it starts to matter for recursive data structures.

There's also the problem that if you limit yourself to simply-typed lambda calculus (i.e. no `forall`-types) and don't introduce any base types, there isn't a single well-typed term in the whole language.

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,711 @@
...
+In this case let's consider an $\mathbf{if} \ldots \mathbf{then} \ldots \mathbf{else} \ldots$ operator.
+
+The semantics of this operator should be reasonably familiar to a practicing programmer, but let's informally write them them out anyway:
+
+- $\mathbf{if}\; true\; \mathbf{then}\;\alpha\;\mathbf{else} \;\beta$ should reduce to $\alpha,$
+- $\mathbf{if}\; false\; \mathbf{then}\;\alpha\;\mathbf{else} \;\beta$ should reduce to $\beta,$
+- $\mathbf{if}\; \gamma\; \mathbf{then}\;\alpha\;\mathbf{else} \;\beta$ where $\gamma$ is a redex, $\gamma$ should be reduced first.
+
+The introduction of the $\mathbf{if}$ operator creates a bit of a kink.
+What do we do if $\gamma$ is neither $true$, $false$, nor a redex?
```
int-index, line 49: Could you give an example of that? E.g.

```
if (\x -> x) then ... else ...
```

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,711 @@
...
+
+$$
+\begin{array}{c}
+\text{premises} \\
+\hline
+\text{conclusion}
+\end{array}
+$$
+
+where $\text{premises}$ is zero or more semicolon-separated typing relations, and $\text{conclusion}$, also often called judgment, is a single typing relation.
```
int-index: The terminology seems a bit off. My understanding is that
* A _typing judgment_ is a single judgment about some term and some type, e.g. `true : Bool` or `(λx. x) : (Bool → Bool)`. 
* The _typing relation_ is the set of all typing judgments `a : t` in a given type system.
* _Typing rules_ define how to derive a typing judgment (conclusion) from other typing judgments (premises).

Therefore, the premises and the conclusion of a typing rule contain typing judgments, not typing relations. There's just one typing relation in a type system, and it is defined by the typing rules.

(A type system may also contain other relations. E.g. if we were to formalize Haskell, we'd have to define the typing relation `a :: t` and the kinding relation `t :: k`. Or the typing relation could actually be a family of relations parameterized by something, e.g. up-or-down direction in bidirectional typechecking. But that's not so important here)

With that in mind, I suggest the following correction:

```suggestion
where $\text{premises}$ is zero or more semicolon-separated typing judgments, and $\text{conclusion}$ is a single typing judgment.
```

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,711 @@
...
+false : Bool & \text{(AxF)}
+\end{array}
+$$
+
+And for the $\mathbf{if}$ operator:
+
+$$
+\begin{array}{cc}
+\begin{array}{c}
+t_1 : Bool;\quad t_2 : T;\quad t_3 : T \\
```
int-index, line 109: Is there a place where I could take a look at the rendered markup?

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,711 @@
...
+The exact features of the type system used can be chosen mostly freely, and we will discuss some common choices in this article.
+
+The untyped lambda calculus is hard to use as a logic system due to the paradoxes it creates.
+Adding Russell's type theory to lambda calculus makes it, under certain conditions, logically consistent, which in turn makes it possible to mathematically guarantee certain behaviors (or lack thereof).
+
+The most common argument for the strongly-typed languages is that if a program type checks, it will never segfault (or have any other type-related issues), but I think this argument sells typed languages short.
+A sufficiently strict and expressive type system allows for encoding much more than simply "not segfaulting".
+In some cases, the whole domain can be expressed in types.
+
+In this article, we're concerned with the basics, for the most part, so we won't get too much into the more expressive type systems like System FC or systems with dependent types.
```
int-index: Thanks for updating the intro! I believe it's much better now

typed-lambda-calculus/article.md
lierdakil: https://hackmd.io/@lierdakil/HJMIrLLX6

typed-lambda-calculus/article.md
lierdakil: Yes, I'm intentionally sweeping a lot under the rug here. I discuss how pure STLC is impossible later in the article, but I can't really discuss it without introducing STLC first, and this paragraph is supposed to provide _some_ motivation, however flimsy, for introducing atomic values without getting into the weeds. The weeds are for later.

Re: computational complexity, this theoretical introduction is far removed from the practicalities of silly (nb: sarcasm) things like computaional complexity of common operations, so I reckon it's perfectly reasonable to call such things "practically not very convenient" here. If you believe it's important to note, I can add a blurb in a cutout somewhere near, briefly discussing why atomic values are good for performance.

typed-lambda-calculus/article.md
lierdakil: Done.

typed-lambda-calculus/article.md
lierdakil: Ahh, right, thanks for the correction. Some wires apparently got crossed in my head, I blame faulty memory. I did my best to amend this, hopefully to a satisfactory degree, but if you notice wonky terminology further in the article please let me know.

typed-lambda-calculus/article.md
lierdakil: Side note, you should be able leave comments there, so feel free to do so if it's more convenient. Apparently nobody cares about this repo anymore.

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,717 @@
...
+
+Such terms are called "stuck", the computation, according to our rules, can not proceed.
+We want to avoid stuck terms!
+
+To do that, we'll have to place some restrictions on what $\gamma$ could be.
+
+### Typing relations, rules, and systems
+
+To express such restrictions, we introduce the so-called _typing relation_.
+A typing relation is a relation, in this case meaning a set of pairs, which assigns a _type_ to each and every expression in the language.
```
int-index: When you say "each and every expression in the language", I suppose the idea is that only well-typed terms are "in the language", but then the statement becomes somewhat circular.

It would be more clear to consider all syntactically valid terms. Then you could say that the typing relation assigns types to _some_ expressions in the language, and those expressions are considered _well-typed_. At the same time, expressions that do not have a type assigned to them by the typing relation are _ill-typed_.

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,717 @@
...
+
+One such individual assignment we'll call _typing judgement_, and will write it separated by a column: $t : T$, read as "term $t$ has type $T$".
+
+Types can be thought of as sets of terms of a certain general form.
+Turns out (Reynolds, 1984), this naive interpretation of types as sets doesn't hold to rigorous scrutiny.
+Still, it's a good enough mental model unless you're a type theory researcher (and if you are, hi, very nice of you to read this article, please get in touch if you have corrections).
+
+Typing relation thus assigns a type to each term.
```
int-index: Same comment as above. It's not each term, it's only some of them (and those are well-typed). All other terms are ill-typed.

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,717 @@
...
+We should note that the $\text{(If)}$ typing rule will forbid some "correct" terms, as in "terms that aren't stuck".
+For example, $\mathbf{if}\;true\;\mathbf{then}\;true\;\mathbf{else}\;(\lambda x.\; x)$ is just $true$, which is arguably perfectly reasonable, but our type system wouldn't allow it.
+
+This is true in general: any sufficiently strict type system will reject some "correct" terms, but that's basically a consequence of Gödel's incompleteness theorems, so either we have to deal with a system that allows for paradoxes (inconsistency), or we have to live with some arguably correct terms being rejected (incompleteness).
+
+In this particular example, it makes sense if you think about it.
+The example above is trivial enough, but consider you might have $\mathbf{if}\;\text{<long and hard computation>}\;\mathbf{then}\;true\;\mathbf{else}\;(\lambda x.\; x),$ and this becomes much less clear what the type should be without executing the program.
+That is not to say this is a fundamental restriction, more powerful type systems can lift it, but we won't discuss them here.
+
+We call a type system _sound_ if it assigns (relates) a single type to each term.
```
int-index: at most one

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,717 @@
...
+
+$$
+\begin{array}{c}
+\Gamma \vdash (\lambda x:Bool.\; x) : T_1 \to T_2;\quad \Gamma \vdash true : T_1 \\
+\hline
+\Gamma \vdash (\lambda x:Bool.\; x)\; true : T_2
+\end{array} \quad\text{(App)}
+$$
+
+For each condition, we do the same in turn:
```
int-index: I'm not sure typing trees work this way. What you described looks like an informally-specified type inference algorithm, but as far as typing trees themselves are concerned, you're simply supposed to magically know the right answers.

So I agree with the typing tree you give at the end of this section, whereas the intermediate result is not a valid typing tree because it's not valid to selectively instantiate parts of it. It looks like an intermediate data structure that could be used by an inference algorithm.

I guess my problem is only with presentation. It is easy to read this section and end up with the misconception that there's a canonical method to construct typing trees. In reality, there might be more than one algorithm for a given type system, or the type system could be undecidable. To avoid discussing any algorithms, you can assume the existence of an oracle that just magically tells you the type of a term, and then building the typing tree merely validates that the oracle didn't lie to you. (In a compiler, we have the type inference engine instead of the oracle)

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,717 @@
...
+But we can't do it in general.
+
+Notice how we need a terminating type in the type signature.
+Consequently, if our calculus didn't have any atomic values, i.e. didn't have any types beyond abstractions, we wouldn't be able to specify the argument type.
+
+The reason for this is the lack of polymorphism.
+
+## Polymorphic lambda-calculi
+
+Polymorphism, essentially, is the ability of the type system to represent terms that contain arbitrary types.
```
int-index: In what sense do terms "contain" types? Maybe you're looking for a different verb

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,717 @@
...
+Parentheses can be used in value and type expressions to change evaluation order, but we omit these in the syntax definitions for brevity.
+
+Note that type variables are admitted as monotypes, which makes those distinct from _monomorphic types_ that can't have type variables.
+
+Polytypes also include monotypes, in addition to quantified types.
+
+The quantifier $\forall$ _binds_ the type variables, so in a type expression $\forall \alpha_1\;\ldots\;\alpha_n.\; \tau,$ all of $\alpha_i$ are _bound_.
+Type variables that are not bound are called _free_.
+
+Additionally, type variables may be bound by the typing context.
```
int-index: This sounds a bit odd to me. The typing context exists at a meta-level, whereas I'd expect variable bindings to occur directly in the syntax. However, I'm not very familiar with HM. I suppose I just don't know enough about HM to review this section of the article – saying this just for the record.

typed-lambda-calculus/article.md
```diff
@@ -0,0 +1,717 @@
...
+The most general type $\forall\alpha.\;\alpha$ is usually considered uninhabited (with a few caveats), i.e. it has no values at all.
+
+Formally, this all can be expressed by the following rule (called specialization rule):
+
+$$
+\begin{array}{c}
+\beta_i \notin free(\forall\alpha_1\ldots\alpha_n.\;\tau) \\
+\hline
+\forall\alpha_1\ldots\alpha_n.\;\tau
+\sqsubseteq \forall\beta_1\ldots\beta_m.\;\{\alpha_i\mapsto\tau_i\}\tau
```
int-index, line 504: I was initially unsure how to read this rule, but then I realized that it assumes that `tau_i` draws from a set of types `tau_1 .. tau_n`, all of which can mention type variables from `beta_1 .. beta_m`. This is implicit in the rule and could be spelled out in an accompanying paragraph.

typed-lambda-calculus/article.md
int-index: System F is introduced later in the article and it binds type variables the way I expected (with a big lambda directly in the syntax). Apparently my expectations were shaped by the way GHC does things internally, and it's based on System F, so that makes sense.

lierdakil: You're right, it does sound weird. Poor choice of words on my part. It would be more accurate to say that type variables are considered bound if they occur in the context.

int-index:
This was published at https://serokell.io/blog/look-at-typed-lambda-calculus

PR #27, opened 2022-11-01
[PR-1498] Article about the Alternative type class

alios:
Add links to other articles

### DIFF
--- a/whats-that-typeclass-alternative/article.md
+++ b/whats-that-typeclass-alternative/article.md
@@ -1,0 +1,446 @@
+You may come across a programming task where you want to implement something like:
+
+Try _this_ and, if it doesn't work out, try _that_ - or one could say a computation composed of two (or more) alternatives which should be tried
+one after the other until one of them produces a useful result.
+
+As an example, lets build a function which takes a `String` as input and returns `Just String` if that input is either `"bob"` or `"alice"`.
+
+Although there might be a lot of other ways to solve this task, we'll do it by defining two functions:
+
+```haskell
+isAlice :: String -> Maybe String
+isAlice a = if a == "alice" then Just a else Nothing
+
+isBob :: String -> Maybe String
+isBob a = if a == "bob" then Just a else Nothing
+```
+
+We can then compose these two functions like this:
+```haskell
+isAliceOrBob :: String -> Maybe String
+isAliceOrBob a =
+  case isAlice a of
+    Just _ -> Just a
+    Nothing -> case isBob a of
+                 Just _ -> Just a
+                 Nothing -> Nothing
+```
+
+```none
+λ> isAliceOrBob "alice"
+Just "alice"
+λ> isAliceOrBob "bob"
+Just "bob"
+λ> isAliceOrBob "charly"
+Nothing
+```
+
+This works well, but it will become ugly when we add more alternatives to the computation  f.e. a `isCharly :: String -> Maybe String`.
+
+It would be nice if we could have an operator to write it in a nice way like this:
+
+```haskell
+isCharly :: String -> Maybe String
+isCharly a = if a == "charly" then Just a else Nothing
+
+isAliceOrBobOrCharly :: String -> Maybe String
+isAliceOrBobOrCharly a = isAlice a <|> isBob a <|> isCharly a
+-- or
+isAliceOrBobOrCharly a = isAliceOrBob a <|> isCharly a
+```
+
+
+## The Alternative typeclass
+
+Actually, the code above is valid Haskell. All you need to do to make use of the `<|>` operator is to import it from `Control.Aplicative`:
+
+```none
+λ> :info (<|>)
+type Alternative :: (* -> *) -> Constraint
+class Applicative f => Alternative f where
+  ...
+  (<|>) :: f a -> f a -> f a
+  ...
+  	-- Defined in ‘GHC.Base’
+infixl 3 <|>
+```
+
+It's a method of the `Alternative` typeclass. Let's look at what information GHCI can provide us about this typeclass:
+
+```haskell
+type Alternative :: (* -> *) -> Constraint
+class Applicative f => Alternative f where
+  empty :: f a
+  (<|>) :: f a -> f a -> f a
+  some :: f a -> f [a]
+  many :: f a -> f [a]
+  {-# MINIMAL empty, (<|>) #-}
+  	-- Defined in ‘GHC.Base’
+instance Alternative ZipList -- Defined in ‘Control.Applicative’
+instance Alternative [] -- Defined in ‘GHC.Base’
+instance Alternative Maybe -- Defined in ‘GHC.Base’
+instance Alternative IO -- Defined in ‘GHC.Base’
+```
+
+This already gives us a lot of useful facts:
+
+1. A type implementing `Alternative` must be of [kind](https://serokell.io/blog/kinds-and-hkts-in-haskell) `* -> *`.
+2. It must also implement `Applicative` (and, with that, also `Functor`).
+3. To implement it, you must implement `<|>` and `empty`.
+4. In addition to those methods, it gives you `some` and `many`, which we'll discuss further below.
+5. It's already defined for the `Maybe` type that we used in the example above.
+
+Lets take a look on the implementation of `Alternative Maybe`:
+
+```haskell
+instance Alternative Maybe where
+    empty = Nothing
+    Nothing <|> r = r
+    l       <|> _ = l
+```
+
+If you take a look at the definitions you can see that `empty :: f a` is
+the identity of `<|>` (very much like `mempty :: a` is the identity of
+`mappend :: a -> a -> a` in case of `Monoid`.) `Alternative` is also
+said to be _a monoid on applicative functors_.
+
+The truth table of `Alternative Maybe` looks like this:
+
+|             | **Nothing** | **Just r** |
+|-------------|-------------|------------|
+| **Nothing** | Nothing     | Just r     |
+| **Just l**  | Just l      | Just l     |
+
+
+Another interesting `Alternative` instance is `IO`.
+Here, the first alternative from left to right that succeeds wins, f.e.
+
+```haskell
+test :: IO String
+test = fail "foo" <|> pure "bar" <|> fail "baz"
+```
+
+When executed the first computation `fail "foo"` obviously fails, so the second one `pure "bar"` succeeds, which is why the third `fail "baz"` isn't executed at all.
+
+```none
+λ> test
+"bar"
+```
+
+## A simple parser using Alternative
+
+A very common use case of the `Alternative` typeclass is in parsers.
+
+Lets assume we want to implement a simple parser which reads a stream of 'i' and possibly returns a value of type `a` if the input matches.
+
+```haskell
+newtype Parser i a
+  Parser { runParser :: [i] -> (Maybe a, [i]) }
+
+evalParser :: (Show i) => Parser i a -> [i] -> Either String a
+evalParser p is = case runParser p is of
+                    (Nothing, is) -> Left $ "parser failed on " ++ show is
+                    (Just a, _)   -> Right a
+```
+
+This `Parser` type wraps a function which takes an input and gives back `Just a` in case it matches and `Nothing` if it doesn't, together with the remaining unused input.
+
+`evalParser` runs such a parser and returns a more convenient `Either` with
+an error string or the result.
+
+
+An implementation of the parser could be one which consumes a numeric input
+and succeeds if it smaller then a given number of fails if it is equal
+or greater that number:
+
+
+```haskell
+parseLT :: Ord a => a -> Parser a a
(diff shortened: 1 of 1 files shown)


whats-that-typeclass-alternative/article.md
```diff
@@ -0,0 +1,398 @@
...
+λ> evalParser (parseLT 10) []
+Left "parser failed on []"
+```
+
+### Implementing a Alternative for our parser
+
+
+First we need a `Functor` instance for our `Parser i a` type which is pretty
+straight forward reusing the `Functor` instance for our inner `Maybe a`.
+
```
alios, line 185: Link the functor whats-that-type-class article.

whats-that-typeclass-alternative/article.md
```diff
@@ -0,0 +1,398 @@
...
+    (Parser f) <*> (Parser p) = Parser $ \is ->
+      let (fM, rest) = f is
+      in case fM of
+           Nothing -> (Nothing, rest)
+           Just f' -> case p rest of
+                        (Just o, rest')  -> (pure $ f' o, rest')
+                        (Nothing, rest') -> (Nothing, rest')
+```
+
+I won't go into the details of `Applicative` here, as there are excelent other
```
alios: Link the applicative whats-that-type-class article.

whats-that-typeclass-alternative/article.md
```diff
@@ -0,0 +1,398 @@
...
+instance Alternative Maybe -- Defined in ‘GHC.Base’
+instance Alternative IO -- Defined in ‘GHC.Base’
+```
+
+This already gives us alot of useful information about that typeclass:
+
+1. A type implementing `Alternative` must be of kind `* -> *`
+2. It must also implement `Applicative` (and with that also `Functor`)
+3. To implement it you must implement `<|>` and `empty`
+4. Next to those 2 functions it gives you `some` and `many` as well
```
NaeosPsy: You don't really cover what these do, I think it would be beneficial.

whats-that-typeclass-alternative/article.md
```diff
@@ -0,0 +1,398 @@
...
+Right [2,3]
+
+λ> evalParser (some $ parseLT 10) [10,2,3,20]
+Left "parser failed on [10,2,3,20]"
+
+λ> evalParser (many $ parseLT 10) [10,2,3,20]
+Right []
+```
+
+`some` and `many` are a very common and useful pattern in parser combinators.
```
NaeosPsy, line 275: Can add a link to: https://serokell.io/blog/parser-combinators-in-haskell

whats-that-typeclass-alternative/article.md
```diff
@@ -0,0 +1,398 @@
...
+```none
+λ> isAliceOrBob "alice"
+Just "alice"
+λ> isAliceOrBob "bob"
+Just "bob"
+λ> isAliceOrBob "charly"
+Nothing
+```
+
+This works well but will become ugly soon when we want to add more alternatives to the computation  f.e. a `isCharliy :: String -> Maybe String`.
```
NaeosPsy: isCharly, not isCharliy

whats-that-typeclass-alternative/article.md
```diff
@@ -0,0 +1,398 @@
...
+
+An implementation of a parser could be one which consumes a numeric input
+and succeeds if it smaller then a given number of fails if it is equal
+or greater that number:
+
+
+```haskell
+parseLT :: Ord a => a -> Parser a a
+parseLT a = Parser f
+    where f input@(i : is)
```
NaeosPsy, line 160: This is a type application, right? At what point in their Haskell journey do we expect people to just know it? 😅

NaeosPsy: Given that InstanceSigs is explained, it seems odd to not explain this one, if I understand correctly.

whats-that-typeclass-alternative/article.md
alios: done

whats-that-typeclass-alternative/article.md
alios: done

whats-that-typeclass-alternative/article.md
alios: added a note that I will explain that further below

whats-that-typeclass-alternative/article.md
alios: done

whats-that-typeclass-alternative/article.md
alios: You mean the `@`? 

No, thats not a type application but the name for the whole list (while `i` refers to the first element and `is` to the remaining so `input = i : is`.

Use that here because I want to:
1. Pattern match on the non empty list 
2. Have the `i` at hand for the 1st guard (` i < a     = (Just i, is)`)
3. But In the other guard case I need the whole list (`| otherwise = (Nothing, input)`). 
    Without the `input@` I could have written it as `| otherwise = (Nothing, i:is)`
    
I guess thats pretty standard haskell syntax which does not require any language extension.

NaeosPsy: Ah, "as pattern". Cool, learned something new today, thanks!

whats-that-typeclass-alternative/article.md
```diff
@@ -0,0 +1,414 @@
...
+  	-- Defined in ‘GHC.Base’
+instance Alternative ZipList -- Defined in ‘Control.Applicative’
+instance Alternative [] -- Defined in ‘GHC.Base’
+instance Alternative Maybe -- Defined in ‘GHC.Base’
+instance Alternative IO -- Defined in ‘GHC.Base’
+```
+
+This already gives us a lot of useful facts:
+
+1. A type implementing `Alternative` must be of kind `* -> *`.
```
heitor-lassarote: Is this article aimed at beginners? It's very likely they don't know what `* -> *` is. We have an article about kinds, might be nice to add a link to it.

whats-that-typeclass-alternative/article.md
```diff
@@ -0,0 +1,414 @@
...
+
+Another interesting `Alternative` instance is `IO`.
+Here, the first alternative from left to right that succeeds wins, f.e.
+
+```haskell
+test :: IO String
+test = fail "foo" <|> pure "bar" <|> fail "baz"
+```
+
+When executed the fist computation `fail "foo"` obviously fails, so the second one `pure "bar"` succeeds, which is why the third `fail "baz"` isn't executed at all.
```
heitor-lassarote: s/fist/first/

whats-that-typeclass-alternative/article.md
```diff
@@ -0,0 +1,414 @@
...
+λ> evalParser (some $ parseLT 10) [10,2,3,20]
+Left "parser failed on [10,2,3,20]"
+
+λ> evalParser (many $ parseLT 10) [10,2,3,20]
+Right []
+```
+
+`some` and `many` are a very common and useful pattern in parser combinators.
+
+If you want to dive deeper into the topic of parsers in Haskell, take a look
```
heitor-lassarote, line 277: I'd make it evident that it's about parser combinators, as we have other blog posts about parsers too.

BTW, what do you think about changing this blog post to include references to this one too?

whats-that-typeclass-alternative/article.md
```diff
@@ -0,0 +1,414 @@
...
+* `Alternative` is a useful instance to implement for your applicative functor
+  if it has a semantic of _try this or alternatively that_.
+* `Alternative` can be seen as a monoid on applicative functors.
+* If you implement a parser (combinator), you almost always want to implement
+  an `Alternative` instance for it.
+
+Another very closely related typeclass is [`MonadPlus m`](https://wiki.haskell.org/MonadPlus)
+which has the same semantics as `Alternative` but for types which also implement `Monad`. 
+So it can be seen as a "monoid on monads". It comes with a default implementation
+based on `Alternative`, so you don't have to actually implement any methods:
```
heitor-lassarote: Could you explain when one would choose `Alternative` over `MonadPlus` or vice-versa?

whats-that-typeclass-alternative/article.md
```diff
@@ -0,0 +1,414 @@
...
+`some` and `many` are a very common and useful pattern in parser combinators.
+
+If you want to dive deeper into the topic of parsers in Haskell, take a look
+[at this post](https://serokell.io/blog/parser-combinators-in-haskell).
+
+## Conclusion and outlook
+
+* `Alternative` is a useful instance to implement for your applicative functor
+  if it has a semantic of _try this or alternatively that_.
+* `Alternative` can be seen as a monoid on applicative functors.
```
heitor-lassarote: Note that there is `Data.Monoid.Alt` which further shows this idea.

whats-that-typeclass-alternative/article.md
```diff
@@ -0,0 +1,414 @@
...
+   our new `Parser i a` type.
+
+   So you come up with a parser:
+
+   ```haskell
+   parseAliceBobCarly :: Parser Char String
+   parseAliceBobCarly = ...
+   ```
+
+   You are allowed to utilize functions from `Data.List`
```
heitor-lassarote, line 351: Do you think it's worth it to talk about `Data.Foldable.asum`? It's a useful thing that you frequently run across.

whats-that-typeclass-alternative/article.md
NaeosPsy: I think you can expect that the people reading this article are a bit more experienced and know what a kind is. 

(It's always nice to cross-link things, though. https://serokell.io/blog/kinds-and-hkts-in-haskell <-- here's the link)

whats-that-typeclass-alternative/article.md
NaeosPsy: We can add references from it once we publish the Applicative one and this, yeah. 👍

whats-that-typeclass-alternative/article.md
alios: added a link the article.

whats-that-typeclass-alternative/article.md
alios: although I'd like the idea of reasoning about what a "fist computation` might be, I fixed it ;)

whats-that-typeclass-alternative/article.md
alios: The parser article is linked some lines below, but I like the idea of linking back!

whats-that-typeclass-alternative/article.md
alios: Added a paragraph about the "more powerfull"  monadic context by the example of `mfilter`.

whats-that-typeclass-alternative/article.md
alios: Good idea, added and linked it.

whats-that-typeclass-alternative/article.md
alios: Yes it is, linked it and provided an alternative implemenation using `asum`

whats-that-typeclass-alternative/article.md
```diff
@@ -0,0 +1,446 @@
...
+   -- @
+   -- mplus = ('<|>')
+   -- @
+   mplus :: m a -> m a -> m a
+   mplus = (<|>)
+```
+
+Altough you could use the `Alternative` machinery everywhere where you could use `MonadPlus`
+(as every `Monad` is also an `Applicative`) there are useful functions which only work in the
+context of a `MonadPlus` f.e. [`mfilter :: MonadPlus m => (a -> Bool) -> m a -> m a`](https://hackage.haskell.org/package/base-4.17.0.0/docs/Control-Monad.html#t%3AMonadPlus) which filters a `MonadPlus` with a predicate:
```
heitor-lassarote: Here you linked to `MonadPlus`, did you meant to link to `mfilter` instead?

whats-that-typeclass-alternative/article.md
```diff
@@ -0,0 +1,446 @@
...
+`some` and `many` are a very common and useful pattern in parser combinators.
+
+If you want to dive deeper into the topic of parsers in Haskell, take a look
+[at this post](https://serokell.io/blog/parser-combinators-in-haskell).
+
+## Conclusion and outlook
+
+* `Alternative` is a useful instance to implement for your applicative functor
+  if it has a semantic of _try this or alternatively that_.
+* `Alternative` can be seen as a monoid on applicative functors.In the
```
heitor-lassarote: Nitpick: you missed a space before the dot.

heitor-lassarote (approved):
The article looks quite good, thanks for addressing my comments. Left just a few final nits, but I'm already approving it.

whats-that-typeclass-alternative/article.md
alios: oh thnx - fixed

whats-that-typeclass-alternative/article.md
alios: fixed as well, thnx

PR #28, opened 2022-11-04
WIP: Article about the first phase of the collaboration with Runtime Verification

breakerzirconia:


kframework/article.md
```diff
@@ -0,0 +1,169 @@
...
+       @ c_aM8
+       eta2_B1
+       (\ (b1_anv :: b_aM7) -> eta_B3 a1_anu b1_anv))
+```
+
+The call to `>>=` is **not** inlined, and the reason is because it is instantiated at an abstract monad `m_aM5`, and GHC does not now which instance to use. Instead, it has to pass the implementation in a dictionary `$dMonad_aMr`.
+
+## Findings
+
+Here is a list of relevant terminology, datatypes, and typeclasses:
```
NaeosPsy: Note: it's odd to have a section called Findings that starts listing terminology at the start of it. 😅

breakerzirconia: Right, I guess I need to refactor this part a little.
This list of terminology, datatypes, and typeclasses (let's call it _list1_) is relevant only to the _Specialization_ subsection.  The _Cache hit ratio_ and the _String Interning_ subsections also contain their respective terminologies at the start, which is inconsistent with how _list1_ is placed (**before** the _Specialization_ subsection and after _Findings_).

Which option is more logical?
1. To place _list1_ after the _Specialization_ header?
2. To place all definitions from **all** the aforementioned subsection to another standalone level 2 header (e.g. _Relevant Terminology, Datatypes, and Typeclasses_)?

P.S. I noticed another inconsistency in my draft: the differing capitalization rules in _Cache hit ratio_ and _String Interning_. Which capitalization style is preferred?

NaeosPsy: Sentence case, e.g., `### Cache hit ratio` for headings

NaeosPsy: It's better to have the definition as close to the usage point as possible. So definitely not a big vocabulary as a subsection. 

The way you have it at the start of the each section is fine, I personally prefer for it to be weaved in the text more.

kframework/article.md
```diff
@@ -0,0 +1,169 @@
...
+We attempted to solve the [aforementioned problem of inlining](#inlining) by changing the abstract monad `MonadSimplify simplifier` to `SimplifierT SMT`. This could not be done easily because of the application architecture: the base monad, `SMT` or `NoSMT`, is determined dynamically.
+
+```
+execute options metadataTools lemmas worker =
+  case solver of
+    Z3 -> withZ3
+    None -> withoutSMT
+```
+
+For the purposes of this investigation we removed the support of `NoSMT`.
```
gromakovsky: We didn't completely remove it though. `withoutSMT` still [exists](https://github.com/runtimeverification/haskell-backend/blob/a1959233e210f960d49a91ebe7fcec76da14ebe3/kore/app/share/GlobalMain.hs#L558). We have changed the implementation so that the information where SMT solver is provided is stored inside runtime environment.

kframework/article.md
```diff
@@ -0,0 +1,169 @@
...
+
+Turned out, the `simplifier` type variable is instantiated to different monad transformer stacks and could not be easily monomorphized in some functions. We developed a custom tool, `Debug.Trace.Usage`, in order to determine how the monad is being instantiated in practice. Here is the summary of its results on one of the test cases:
```
gromakovsky: I would upload the code of `Debug.Trace.Usage` into GitHub gist and make it clickable here.

kframework/article.md
```diff
@@ -0,0 +1,169 @@
...
+
+```
+1 MaybeT (ReaderT (SideCondition RewritingVariableName) (UnifierT (LogicT Simplifier)))
+2 MaybeT (ReaderT (SideCondition RewritingVariableName) (UnifierT (LogicT (LogicT Simplifier))))
+10 MaybeT (ReaderT (SideCondition RewritingVariableName) (UnifierT (StateT (MultiAnd StuckClaim) (ExceptT (MultiAnd StuckClaim) (StateT (MultiAnd SomeClaim) Simplifier)))))
+14 MaybeT (ReaderT (SideCondition RewritingVariableName) Simplifier)
+54087 Simplifier
```
gromakovsky, line 148: It's rather not obvious how to interpret what's written here. Especially 54087, what is it?

gromakovsky: So AFAIU these are 5 examples of instantiantions of some type parameters, but what do these numbers mean?

gromakovsky: > The vast majority (99.95%) of calls are in the `Simplifier` monad

Aha, ok, I get it now. So there is, roughly speaking, `f :: m ()` somewhere and in 54087 cases it's instantiated to `Simplifer`, in 14 cases it's `MaybeT (ReaderT (SideCondition RewritingVariableName) Simplifier)`, etc. Right?

Anyway, since it took me some time to understand what's written here (and I am not new to the project, unlike potential readers), I think it would be good to make it a bit clearer.

breakerzirconia: Oh, of course! I will add a statement on how to interpret these lines in the report. Thank you!

> in 14 cases it's MaybeT (ReaderT (SideCondition RewritingVariableName) Simplifier), etc

That's it, yes.

kframework/article.md
breakerzirconia: This is the phrase that I copied from Vladislav's performance report when the discovery phase was conducted back in May 2022. I think you're referring to the ultimate fate of the SMT datatype when we had merged the two by wrapping the environment in the Maybe context.

Perhaps, I can restructure this sentence a little bit, to make sure the reader is able to unambiguously understand that this maneuver is performed specifically in the discovery phase:
_For the purposes of this discovery phase inverstigation,_

What do you think?

kframework/article.md
breakerzirconia: Hmm, can we simply link the already existing serokell [repository](https://github.com/serokell/trace-usage) that contains `Debug.Trace.Usage`?

gromakovsky: Oh, I didn't know/remember about this repo. Yes, in this case we can.

kframework/article.md
gromakovsky: Yes, sounds good to me.

gromakovsky: @breakerzirconia this sentence still needs to be updated.

gromakovsky:
@breakerzirconia the new part you wrote looks good to me. Here are some things I'd like to add:
1. Links to PR(s)/commit(s) where the changes you describe have been implemented. Since `haskell-backend` is public, let's show how exactly we did something.
2. Performance impact of this change. I guess it's supposed to go into the `Results` section which isn't written yet. We can write about the combined result of both performance changes (the one you described and the one Sergey is describing) or we can write about performance impact of each individual change. The latter is probably better.
3. I would also add a short "Conclusion" chapter with just 1-2 paragraphs, summarizing what's written above and announcing the next part of the story that will be published as another blog post.

breakerzirconia:
> 1. Links to PR(s)/commit(s) where the changes you describe have been implemented.

I've been basing the first part of the specialization task on the changes made within one commit, so I will think about where to place the link.

> 2. Performance impact of this change.

I was a actually a little hesitant, because, as Vladislav mentioned in the aforementioned commit, that change was insufficient to improve performance, plus there weren't any benchmark results, so I'm not sure what to put in the Results section.

> 3. I would also add a short "Conclusion" chapter.

Sure! Should I delete the _Summary_ subsubsection if we follow this plan?

gromakovsky:
@breakerzirconia 

> I was a actually a little hesitant, because, as Vladislav mentioned in the aforementioned commit, that change was insufficient to improve performance, plus there weren't any benchmark results, so I'm not sure what to put in the Results section.

Ah, yes, now I remember that. But still, I think the reader may be curious what performance impact this change had, since performance improvements are our ultimate goal. So we can explicitly say that this change itself didn't have any noticeable effect on performance, but it was very useful nevertheless. It can be added to "Summary" for example.

> Sure! Should I delete the Summary subsubsection if we follow this plan?

After Sergey's commit we have 2 _Summary_ subsections: one for "Monomorphization and specialization, pt. 1: SMT and NoSMT" and the other one for "String interning". I think it's good and these "Summary" subsections conclude the respective sections. But additionally, I'd like to have a conclusion for the whole article. For example, if you check out [blog posts on our website](https://serokell.io/blog), you will see that many of them have "Conclusion" at the end (I checked 3 and all 3 had it). You can use them for inspiration.

Also I see we have empty chapter called "Results". I suppose we don't need 2 _Summary_ subsections + _Results_ + _Conclusion_. I would just remove _Results_, since results are already part of _Summary_, and I would add _Conclusion_, essentially renaming _Results_ to _Conclusion_.

kframework/article.md
```diff
@@ -251,4 +251,156 @@ Notice, however, that we now have only one base instance: `MonadSMT SMT`, which
...
+    , idLocation :: !AstLocation
+    }
+```
+This means operations such as `==` or `hashWithSalt` operate over `Text` values, which can be slow.
+
+The [first attempt](https://github.com/runtimeverification/haskell-backend/pull/3122) to solve this problem was to use the [intern](https://hackage.haskell.org/package/intern) package. Modification of `Eq` instance for type `Id` to take advantage of interning resulted in approximately 8% speedup on the integration test suite. 
+
+In our attempt to modify the `Ord` instance to take advantage of interning as well, some of our unit and integration tests failed because they assumed that the `Id`s would be sorted alphabetically.  When we profiled it, we noticed that `Kore.Syntax.Id.compare` doesn't play much of a role there, it isn't one of the most expensive cost centers. Moreover, some data persists between invocations of the backend via serialization, and the ordering of `Id`s may change between runs if they are interned in a different order.
+
+However, we also wanted to compare the intern package solution with our [custom interning solution](https://github.com/runtimeverification/haskell-backend/pull/3217).
```
gromakovsky: Actually the main motivation was not the we wanted to compare our custom solution with the library, but that we wanted to have an easier way to modify or debug the code responsible for string interning. And since the library was quite small, it seemed easier to just implement it in our code.
In particular, there is `cacheWidth` parameter in `intern` and `intern` creates multiple maps to speed-up concurent access. We didn't need that and created a newtype to set `cacheWidth` to 1.

kframework/article.md
```diff
@@ -251,4 +251,156 @@ Notice, however, that we now have only one base instance: `MonadSMT SMT`, which
...
+We then define a datatype for the global cache, which will contain all of the process' interned strings:
+```hs
+data InternedTextCache = InternedTextCache
+    { internedTexts :: !(HashMap Text InternedText)
+    , counter :: {-# UNPACK #-} !Word
+    }
+```
+The `InternalTextChache` has two fields:
+  - `internedTexts` is a `HashMap` that stores all the interned strings
+  - `counter` is an incremental counter used to generate new unique IDs
```
gromakovsky: It is needed because `HashMap.size` is not a constant-time operation.

kframework/article.md
```diff
@@ -264,11 +264,13 @@ data Id = Id
...
 
-However, we also wanted to compare the intern package solution with our [custom interning solution](https://github.com/runtimeverification/haskell-backend/pull/3217).
+However, we wanted to have an easier way to modify or debug the code responsible for string interning. There is a `cacheWidth` parameter in `interned` typeclass of `inern` package, it creates multiple maps to speed up concurrent access. As we didn't need that, we had to create a [newtype](https://github.com/runtimeverification/haskell-backend/blob/75ae33172a2d221a3c8c0aa2434b6aeea3c0bf69/kore/src/Kore/Syntax/Id.hs#L51-L68) and redefine the `intered` instance with `chacheWidth` set to 1.
+
+Moreover, since the `intern` library is quite small, is seemed easier to implement a [custom interning solution](https://github.com/runtimeverification/haskell-backend/pull/3217) in our code.
```
gromakovsky: Yep, I think it looks good and more accurate this way.

gromakovsky:
After the latest commits Sergey's part LGTM.
What's left:
1. https://github.com/serokell/blog-posts/pull/28#discussion_r1065851422
2. Link to the PR/commit where the first part of monomorphization was implemented.
3. Say explicitly that the first part of monomorphization didn't have any significant performance impact, but was useful nevertheless.
4. Write a conclusion (this one is tricky since it's not obvious how to write it).

kframework/article.md
```diff
@@ -0,0 +1,416 @@
...
+```
+package *
+  profiling: True
+
+package kore
+  profiling-detail: toplevel-functions
+  optimization: True
+```
+
+The resulting executable will be slower than the executable built without the support of profiling, because
```
NaeosPsy: s/support of profiling/profiling support ?

kframework/article.md
```diff
@@ -0,0 +1,416 @@
...
+* A **cache miss** is the opposite of a cache hit, i.e. a situation when the requested data is not contained in the cache.
+* A **cache hit ratio** is the number of cache hits divided by the total number of requests (in other words, the number of cache hits + the number of cache misses).
+
+We found that the cache in `attemptEquation` enjoys the hit ratio of over 95%, so it is well-placed. However, one of the functions that performs the lookup of the `Simplifier` cache, which is essentially a `HashMap`, is expensive, because a lot of time is spent on hashing and comparing keys.
+
+We can either use a different data structure than a `HashMap` to avoid hashing and comparisons entirely, or try to minimize the amount of time required to hash and compare keys by micro-optimizing the respective operations.
+
+### String interning
+
+* **String Interning** is a method of storing exactly one copy of a distinct string value. In general, **interning** is *reusing* objects that are intensionally equal instead of *creating* new ones. The objects must be immutable.
```
NaeosPsy: Will people know what intensionally equal means?

kframework/article.md
```diff
@@ -0,0 +1,416 @@
...
+
+If the argument is a `Just` wrapping a `SolverSetup` value, this datatype is isomorphic to the original `SMT`; otherwise, it's isomorphic to `NoSMT`.
+
+For the sake of convenience, when referring to the original definition of `SMT`, let's use the name `OldSMT`.
+
+### `MonadSMT`
+
+* `MonadSMT m` is a typeclass that allows accessing `SMT` through monad transformers. It had two base instances: `MonadSMT OldSMT` and `MonadSMT NoSMT`.
+
+We now need to construct the *only* base instance, `MonadSMT SMT`, having got the knowledge of how the functions were defined in the original two base instances for the original two datatypes.
```
NaeosPsy: s/, having got/ with the?

kframework/article.md
```diff
@@ -0,0 +1,416 @@
...
+      f = liftSMT . fSMT
+      ```
+      which are defined in terms of `liftSMT`.
+3. `withSolver :: m a -> m a` is the only function whose argument is `m a`, therefore making it impossible to define it via `liftSMT`, so it remains in the typeclass.
+
+### Summary
+
+The unification of `OldSMT` (originally referred to as `SMT` before defining the combined datatype) and `NoSMT` is the first step of the monomorphization and specialization task. Along the way, we have simplified the definition of the `MonadSMT` typeclass by introducing `liftSMT :: SMT a -> m a` and defining the majority of the functions in terms of it, bringing them outside of the typeclass' scope. The link to the pull request containing the implementation of the `SMT`/`NoSMT` fusion is available [here](https://github.com/runtimeverification/haskell-backend/pull/3156).
+
+## String interning
```
NaeosPsy: Doesn't mention what string interning is, perhaps you would like to note it given that we did elsewhere.

kframework/article.md
```diff
@@ -0,0 +1,416 @@
...
+instance Ord InternedText where
+    a `compare` b
+        | internedId a == internedId b = EQ
+        | otherwise = internedText a `compare` internedText b
+    {-# INLINE compare #-}
+```
+
+The `Eq` and `Hashable` instances are straight forward – we use the ID of the interned string for equality comparison and hashing operations.
+
+Our `Ord` instance quickly checks if the IDs of interned strings are equal, if not, we use lexical order to compare the strings themselves.
```
NaeosPsy:
```suggestion
Our `Ord` instance quickly checks if the IDs of interned strings are equal. If they are not, we use lexical order to compare the strings themselves.
```

kframework/article.md
breakerzirconia: Hmm, the people actually might not. I'll remove this word, since in this case the meaning will not change.

kframework/article.md
```diff
@@ -1,12 +1,16 @@
-In this blog post we will be exploring, describing, and dissecting the first phase of the collaboration between Runtime Verification and Serokell! The collaboration involved the work on making optimizations to [K](https://kframework.org/), a rewrite-based executable semantic framework in which programming languages, type systems and formal analysis tools can be defined with the help of configurations and rewrite rules. The blog post will contain the details and the results of the discovery phase, the recommendations which are introduced in the Findings section, and the implementation of 2 global tasks, alongside their performance results.
+In this blog post, we will be exploring, describing, and dissecting the first phase of the collaboration between Runtime Verification and Serokell.
```
breakerzirconia: There are now lots of instances of one trailing space being present at the end of some of the lines. Are they supposed to be there? If not, can I remove them?

kframework/article.md
breakerzirconia: Oh, I have defined what string interning is in the Findings section. Maybe I can add the following line right after the declaration of `## String interning`:
```
Let's recall the definition of string interning.

[[place the definition of string interning here]]
```

kframework/article.md
NaeosPsy: No, you can remove them. :)

kframework/article.md
NaeosPsy: Oh, indeed haha. 😅

No, that's fine I think.

NaeosPsy (approved):

ehildenb:
@NaeosPsy and @gromakovsky one piece of quick feedback: This article focuses a lot on teh performance profiling and such, which I think is valuable. But there has also been a big emphasis on improving the readability and simplicity of the code, with [these guidelines in mind](https://docs.google.com/document/d/1brNoaHuTDMtXkSvF17E4eOGOiS-4cN8JJCOXsKFZz7c/edit) based on [this article about coding practices](https://drewdevault.com/2018/07/09/Simple-correct-fast.html).

I think there could be a good story here about "don't get too crazy with your types if you want to (i) have a large but still understandable codebase, and (ii) care about performance".

NaeosPsy:
I like the idea @ehildenb! 

I do think that we should cover it in a different part if we do. It would be better for the readers if each part focused on one aspect.

kframework/article.md
```diff
@@ -0,0 +1,448 @@
...
```
gromakovsky:
```suggestion
In this blog post, we will be exploring, describing, and dissecting the first phase of the collaboration between [Runtime Verification](https://runtimeverification.com/) and Serokell.
```

PR #33, opened 2023-04-10
Second part about collaboration with Runtime Verification

Sereja313:


Optimizing K Framework Part 2/article.md
```diff
@@ -0,0 +1,552 @@
...
+
+## Monomorphization and specialization
+
+We used the `trace-usage` report to identify type variables that could be replaced with concrete types if only one type was instantiated during execution. For those whose type variables were instantiated to multiple different types, we added `SPECIALIZE` pragmas. We found that most functions could be monomorphized, and only a few required specialization. The process of monomorphization and specialization was divided into five pull requests: [part 1](https://github.com/runtimeverification/haskell-backend/pull/3336), [part 2](https://github.com/runtimeverification/haskell-backend/pull/3346), [part 3](https://github.com/runtimeverification/haskell-backend/pull/3375), [part 4](https://github.com/runtimeverification/haskell-backend/pull/3408), and [part 5](https://github.com/runtimeverification/haskell-backend/pull/3408).
```
gromakovsky: Same links for parts 4 and 5.

Optimizing K Framework Part 2/article.md
```diff
@@ -0,0 +1,552 @@
...
+
+During this process, we also encountered cases of nested `MaybeT` and `ReaderT` caused by redundant `maybeT` and `runReaderT` calls, which added an unnecessary transformer layer in the argument. We resolved these issues by removing these redundant calls, which were straightforward solutions. We implemented these fixes in the pull requests titled [Remove nested ReaderT](https://github.com/runtimeverification/haskell-backend/pull/3418) and [Remove nested MaybeT](https://github.com/runtimeverification/haskell-backend/pull/3419).
+
+## Summary
+
+In this step we simplified the types in codebase by introducing a concrete `Simplifier` monad to replace the `SimplifierT` transformer. This was possible because we had unified `SMT` and `NoSMT` in previous article. We also addressed the issues with nested transformers, which allowed us to monomorphize most of the type variables that were constrained by `MonadSimplify`. This simplified the types and resulted in a speedup of approximately 15%.
```
gromakovsky: I would add that we also added `SPECIALIZE` pragmas where monomorphization wasn't possible.

Optimizing K Framework Part 2/article.md
```diff
@@ -0,0 +1,552 @@
...
+    hashWithSalt salt (Recursive.project -> TermAttributes{termHash} :< _) =
+        salt `hashWithSalt` termHash
+    {-# INLINE hashWithSalt #-}
+
+```
+
+To produce the pre-computed hash value, we use the derived `Hashable` instance of `TermLikeF`, which will call hash on the respective sub-structure wrapped into one of the `TermLikeF` constructors. For all such sub-structures, the derived `Hashable` instance will then recurse into the contents, using the cached hashes directly instead of performing a deep recursion.
+
+Although the change resulted in a 6% speedup, it also caused some integration tests to fail, and unfortunately, we were unable to find a way to fix them, while keeping pre-computed hash in `TermAttributes`.
+So, to avoid storing pre-computed hash in `TermAttributes`, we [cached the hash](https://github.com/runtimeverification/haskell-backend/pull/3338) of the `TermLikeF` in the `TermLike` itself and extract it in the `Hashable` instance instead of recursively traversing into the `TermLikeF`. To achieve this, we modified the `TermLike` definition:
```
gromakovsky:
```suggestion
So, to avoid storing pre-computed hash in `TermAttributes`, we [cached the hash](https://github.com/runtimeverification/haskell-backend/pull/3338) of the `TermLikeF` in the `TermLike` itself and extracted it in the `Hashable` instance instead of recursively traversing into the `TermLikeF`. To achieve this, we modified the `TermLike` definition:
```

gromakovsky (commented):
Looks very good!

PR #36, opened 2023-10-27
[PR-1658] Review findings

karandit:
Problem: I was asked to review the upcoming blogpost written by Sergey

Solution: here are my findings 

- some small typos
- postInstall is just some cosmetic finding, in my personal opinion it is easier to read if a postXXX is after XXX (but now I'm not sure if this postInstall happens after yarnBuild)
- mkdir -p flag ensures that all intermediate folders are created

- an extra finding which is not included in my changes. I'm not sure if the service can (re)start after a reboot. Have you tried it?
here is my concern: https://github.com/serokell/haskell-certification/blob/master/module-userlevel.nix#L101
At least this was the case with vault-secrets' secret files.

### DIFF
--- a/Website deployment for Runtime Verification/article.md
+++ b/Website deployment for Runtime Verification/article.md
@@ -5,7 +5,7 @@ In this blog post, we will provide insights into the website deployment workflow
 Our collaboration focused on creating nix-based deployment workflow for small websites associated with Runtime Verification projects, including:
   - [Kontrol](https://kontrol.runtimeverification.com/): developer tooling for formal verification of smart contracts written in solodity.
   - [ERCx](https://ercx.runtimeverification.com/): developer tooling for ERC token testing.
-  - [Firefly](https://fireflyblockchain.com/): developer tooling for Etherium smart contract testing.
+  - [Firefly](https://fireflyblockchain.com/): developer tooling for Ethereum smart contract testing.
 
 We will walk through a simple example that showcases all the key features of this deployment workflow:
   - Deployment of `master` and `develop` branches using `deploy-rs`.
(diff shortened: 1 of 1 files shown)


Sereja313:
Thank you!

>  postInstall is just some cosmetic finding, in my personal opinion it is easier to read if a postXXX is after XXX (but now I'm not sure if this postInstall happens after yarnBuild)

Yep, `postInstall` actually runs after `yarnBuild`: https://github.com/serokell/nix-npm-buildpackage/blob/991a792bccd611842f6bc1aa99fe80380ad68d44/buildYarnPackage.nix#L103

> an extra finding which is not included in my changes. I'm not sure if the service can (re)start after a reboot.

In this example, we don't need to use the `preStart` script to install a dependency on the service that provides the sercrets file because that service is deployed under the same user that is used to deploy the main service. So we can simply use `after` and `bindsTo`:
```nix
nixosSystem = nixpkgs.lib.nixosSystem {
    system = "x86_64-linux";
    modules = modules ++ [{
      services.hello-world-app = {
        enable = true;
        secretsFile = secretsPathWithSuffix;
        inherit port;
      };
      systemd.services.hello-world-app.preStart = ''
        while ! ( systemctl is-active network-online.target > /dev/null ); do sleep 1; done
      '';
      systemd.services.hello-world-app = {
        after = [ "sops-nix${serviceSuffix}.service" ];
        bindsTo = [ "sops-nix${serviceSuffix}.service" ];
      };
    }];
  };
```
In this case, the presence of a file with secrets will be checked only when `sops-nix-<suffix>.service` completes its work

Sereja313 (approved):

